PR #18, opened 2015-03-20
Fix some 0.4 deprecations and changes

caryan:
Fixes for some of the recent 0.4 changes.

* constructor capitalization
* array concatentation
* rounding vs constructor
* beginswith -> startswith


timholy:
@caryan, thanks for contributing this! It looks like there's some overlap with #19 (sorry, I didn't see this before I started that).

I've gotten to the point of needing this, so I'm going to merge without waiting for Jeff. In some ways yours is more comprehensive, but it breaks support for 0.3, so I'm going to merge the other one. Certainly feel free to complain or make further improvements :-).

caryan:
Yeah, I wasn't careful about backwards compatibility.  I'll close this for now and think about the points you raised in #19.

PR #24, opened 2015-08-10
add __precompile__() to header

stevengj:
since this module is precompile-safe


timholy:
Thanks!

stevengj:
Maybe tag a new release?

timholy:
Good idea, and thanks for the reminder. https://github.com/JuliaLang/METADATA.jl/pull/3052

Issue #25, opened 2015-09-16
8/10/12 bit FixedPoint numbers in the vein of Ufixed


vchuravy:
In JuliaGraphics/ColorTypes.jl#11 the need for FixedPoint numbers with a different bit size from 32bit arises. 

Ufixed is very flexible and nice for representing ColorValues, but it seems to me that its design is a bit different. Ufixed is defined from [0,1] and Fixed32 has a variable range depending on the number of fraction bits. 

For my needs I wouldn't need a variable fraction bit number as all values are within [-1, 1]

PR #27, opened 2015-09-18
[RFC/WIP] Homogenise design

vchuravy:
So I am currently working on reworking the design of FixedPointNumbers so the`UFixed` and `Fixed` work the same way. Currently this conversion is not yet fully done and I would like some early feedback on the idea. (The overall goal is to make #25 possible)

### Type Hierarchy 
````julia
abstract FixedPoint{T <: Integer, B, f} <: Real
immutable UFixed{T<:Unsigned, B, f} <: FixedPoint{T, B, f}
  i :: T
end
immutable Fixed{T <: Signed,B,f} <: FixedPoint{T, B, f}
  i :: T
end
````
- *T* Is the underlying type
- *B* Is the bits available for usage (To make 10 Bit numbers et. possible)
- *f* Is the fractional part of the *B* bit integer

This leads to the fact that the types `Colors` is using are now defined as ``typealias UFixed8  UFixed{UInt8,8,8}` since they are expected to be normalised. That is the place where currently `UFixed` and `Fixed32` feel the most disjoint and it may warrant the introduction of a normalised `FixedPoint` type.

### Rename
In #16 it was proposed to change `Ufixed` to `UFixed` following the changes of Base.

### Fallout
The heaviest user of `FixedPointNumbers` as far as I am aware of is the `Colors` ecosystem. I am willing to help resolve any issue that might come up. Also I don't know how one would deprecate the functionality nicely and it might be easier to create this first in a new package.

### Todo
- [x] Implement fixed point math
- [x] Check compability with current packages
- [x] Add deprecations for type names/ typealiase
- [x] Documentation
- [x] More test cases

The current set of changes is incomplete, since I would rather have feedback before I spend to much time on this.

@TimHoly @JeffBezanson

### CODE REVIEW

**Comment by timholy** [comment]
:+1:

But, do you need the `B` parameter? How would it be used? Currently a `Ufixed12` satisfies `typemax(Ufixed10) = Ufixed10(64.0616)`, because you can use the top 6 bits to represent numbers bigger than 1. I don't know whether you'd think of that as a good thing or a bad thing.

**Comment by vchuravy** [comment]
I think it is a good thing to use those extra bits for  computation, but if I ask a Ufixed for its value there should be a guarantee that it doesn't exceed it type max.
Currently I have not implemented that part yet. (I would do that in the convert functions)

`B` Is so that we are able to say things like: "UFixed{UInt16,10,2} represents a number that has 8 bits for integer part and 2 bits for its fractional part" That is the way `Y'CbCr10` is defined in ITU-R BT.601-7

@timholy I changed the multiplication for `UFixed` numbers to be 

````julia
 UFixed{T,B,f}((Base.widemul(x.i,y.i) + (convert(widen(T), 1) << (f-1) ))>>f,0)
````

Along the lines of how Fixed works. That also guarantees that arithmetic operations always return UFixed types, but the error bounds in the test are then no longer accurate.

**Comment by timholy** [comment]
> but if I ask a Ufixed for its value there should be a guarantee that it doesn't exceed it type max.

Currently the design is to say that those extra bits are perfectly acceptable, and that they are part of its `typemax`. If you want to impose `0 <= value(x) <= 1`, I think that's a new notion. To make it composable, perhaps we want
```jl
immutable BoundedNumber{T,MinVal,MaxVal} <: Number
    value::T
end
```
and then any numeric type can become a `BoundedNumber`.

timholy:
:+1: on the multiplication change.

vchuravy:
Hm I see. So in the original design `UFixed{UInt16, 16, 10}`, `UFixed{UInt16, 12, 10}` and `UFixed{UInt16, 10, 10}` are the same thing? 

I think we could get away with that. That is effectively how this PR is implementing it right now. I see where my miss understanding comes from. I always took `UFixed10` to say: "This is a 10 bit number normalized to 0..1", but actually it means "This is a unsigned 16 bit number with a 10 bit fractional part".

Coming back to my motivating case. `Y'CbCr` requires a (signed/unsigned) 10 bit number with 2 bit for the fractional part or a 12 bit number with 4 bit for the factional part. 

Currently I could just drop the `B` and the PR would still be in good shape and I would have to take care of the clipping/limits in `ColorTypes` and we can talk later about `BoundedNumber`.

timholy:
That would be fine.

vchuravy:
Ok done. Any ideas how to handle the tests?

timholy:
As in, the failing tests? How about wrapping the `Float32` result in `oftype(x*y, <Float32 result>)`?

vchuravy:
@timholy adding the deprecations makes ColorTypes pass locally.

timholy:
Probably easiest is to skip the deprecation and just define `const Ufixed8 = UFixed8` etc.

timholy:
Not sure whether you're ready to merge yet (title might need updating?), but this looks really good to me. Thanks for tackling a much-needed cleanup.

On at least one point I owe you an apology: perhaps could we do a version check `VERSION >= v"0.4.0-rc2"`, and if true use the `@deprecate_binding` mechanism? Sorry to change my mind on this point.

One further thought: on 0.3, `x::Uint8 + y::Uint8` returns a `Uint`. Given that this PR adopts the 0.4 convention, that might be viewed as inconsistent on 0.3. Shall we just bump the minimum required julia version to `0.4-` with this PR?

Just ping again when you're ready to merge.

vchuravy:
I think bumping the minimum required version is a good idea. 

I am currently rewriting the readme and want to add more tests to this. 

An alternative idea I just had is that the implementation of `UFixed` and `Fixed` don't really differ any more and maybe there is no reason for having different types?

We could have one `Fixed` type and replace the macros with `@generated`, but maybe that refactor should happen after merging this.

timholy:
Sounds good. I too noticed that the distinction seemed likely to be irrelevant, but I agree that can happen as a second stage.

I'm going to go ahead and merge #30, as another PR is waiting on it. Hope it doesn't cause you any rebasing headaches.

vchuravy:
@timholy I squashed and rebased everything. The test suite got a bit bigger ;)

README.md
```diff
@@ -14,28 +14,23 @@ This library exports two categories of fixed-point types. Fixed-point types are
...
-To use it, convert numbers to a `Fixed32` type, or call `Fixed32(x)`, which will default
-to constructing a `Fixed32{16}`.
-
-## Ufixed (unsigned fixed-point numbers)
-
-For unsigned integers, there is a family of subtypes of the abstract `Ufixed` type.
 These types, built with `f` fraction bits, map the closed interval [0.0,1.0]
 to the span of numbers with `f` bits.
-For example, the `Ufixed8` type is represented internally by a `Uint8`, and makes
+For example, the `UFixed8` type is represented internally by a `Uint8`, and makes
```
timholy: `UInt8`, not `Uint8`

README.md
```diff
@@ -14,28 +14,23 @@ This library exports two categories of fixed-point types. Fixed-point types are
...
 `0x00` equivalent to `0.0` and `0xff` to `1.0`.
-The types `Ufixed10`, `Ufixed12`, `Ufixed14`, and `Ufixed16` are all based on `Uint16`
+The types `UFixed10`, `UFixed12`, `UFixed14`, and `UFixed16` are all based on `Uint16`
```
timholy: `UInt16`

src/FixedPointNumbers.jl
```diff
@@ -43,9 +42,25 @@ export
...
+realmin{T<: FixedPoint}(::Type{T}) = typemin(T)
+realmax{T<: FixedPoint}(::Type{T}) = typemax(T)
+
+include("fixed.jl")
 include("ufixed.jl")
+include("deprecations.jl")
+
 
+# TODO: rewrite this by @generated
 for T in tuple(Fixed32, UF...)
```
timholy: `Fixed16`?

src/FixedPointNumbers.jl
```diff
@@ -43,9 +42,25 @@ export
...
 reinterpret(x::FixedPoint) = x.i
 
-include("fixed32.jl")
+# comparison
+=={T <: FixedPoint}(x::T, y::T) = x.i == y.i
+ <{T <: FixedPoint}(x::T, y::T) = x.i  < y.i
+<={T <: FixedPoint}(x::T, y::T) = x.i <= y.i
+
+# predicates
+isinteger{T,f}(x::FixedPoint{T,f}) = (x.i&(1<<f-1))
```
timholy: Given the name, this seems like it should return a boolean.

timholy:
Looks great!

The biggest problem is that the tests are timing out on every platform. I think we're going to have to add a section that's something like this (warning: untested) to `.travis.yml`:
```jl
script:
    - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
    - julia -e 'Pkg.clone(pwd()); Pkg.build("FixedPointNumbers")'
    - julia -e 'cd(Pkg.dir("FixedPointNumbers", "test")); include("runtests.jl")'
```
so the tests run with inlining on.

vchuravy:
Let's see if that helps to speed up the travis run, otherwise we will have to reduce the number of tests for `Fixed`.

timholy:
Doesn't seem to be working. I don't understand why---the old tests ran *much* faster on my machine. Or is this a consequence of the new tests? If on your own machine the tests are >10 minutes, that seems a little long.

If it should be better than this, maybe debug with changing that last line of the `script` section to
```
- julia -e 'fn = fieldnames(Base.JLOptions); for f in fn println(f, " => ", getfield(Base.JLOptions(), f)); end; cd(Pkg.dir("FixedPointNumbers", "test")); include("runtests.jl")'
```
This will cause it to dump the current options in a nicely-readable way. (`can_inline` is the one I'm most concerned about, but might as well see them all.)

vchuravy:
The test take about 30min to run on my machine
```julia
INFO: FixedPointNumbers tests passed
1551.492159 seconds (447 allocations: 24.578 KB)
```
I will remove a few of the Fixed tests.

vchuravy:
That seems to have worked. There are still more test cases then there used to be so merge whenever you are ready.

timholy:
Awesome, thanks for your efforts here!

vchuravy:
Thanks for guiding me along :) 

Next step when I find sometime is to unify the code altogether.

PR #28, opened 2015-09-19
Union() -> Union{}

yuyichao:


yuyichao:
OSX downloading error....

timholy:
I restarted the build but got the same error. Note that it applies to both Linux and OSX, anything except 0.3. It's a little worrisome, since other packages don't seem to be having this problem.

yuyichao:
This seems to be the slow down caused by the coverage ...

yuyichao:
Feeding Travis with enough output get rid of the timeout ................................

timholy:
That's an impressively-massive increase in runtime thanks to shutting off inlining and adding coverage testing.

Still, OK to go with it? An alternative is to try fewer cases in the test, or to leave inlining on. I haven't checked, but it seems possible that with @ihnorton's recent improvements in line numbering, we may not need this to get accurate coverage results.

yuyichao:
Forgot to say that I've tested locally and `--inline=no` slow things down by about 10x.

tkelman:
We should write down a list somewhere of packages where enabling/disabling coverage and inlining makes such a noticeable difference to test's run time.

PR #30, opened 2015-09-30
Add unary operator support (-, ~)

timholy:


timholy:
Alarming observation: on Travis, with julia 0.4-rc3 the tests take more than an order of magnitude longer. See https://travis-ci.org/JeffBezanson/FixedPointNumbers.jl/builds/81574138 (from #27). The run on this job is shaping up the same way.

However, on my own machine I get
```
0.3.12-pre+5: 87 s
0.4-rc3: 65 s
```
which is the opposite.

timholy:
I forgot, this is just https://github.com/JeffBezanson/FixedPointNumbers.jl/pull/28#issuecomment-141660467. Nothing to worry about really, but one curious point: its major effect is limited to `fixed32`.

PR #31, opened 2015-10-12
Fix reductions so they don't overflow

timholy:
Base julia widens small integers when performing reductions. This PR implements this behavior for FixedPoint types. Necessitated by #27. CC @vchuravy.


vchuravy:
Looks good to me.

PR #32, opened 2015-10-15
[RFC/WIP] unify the concepts of Fixed and UFixed

vchuravy:
After #27 UFixed and Fixed shared a lot of concepts and they do effectively the same thing, but sometimes a function is only implemented one way.

This is the current status of having only one type of FixedPoint and the underlying Type T defines  whether it is a Unsigned or Signed type.

I added in comments the second implementation if the implementation is ambiguous and I haven't looked at the tests yet.

Merging those two concepts simplifies the code immensely, but it also highlights the subtle differences that where  between Fixed and UFixed.


src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,198 @@ export
...
     scaledual
 
+# T => BaseType
+# f => Number of Bytes reserved for fractional part
+immutable FixedPoint{T <: Integer, f} <: Real
+    i::T
+
+    # constructor for manipulating the representation;
+    # selected by passing an extra dummy argument
+    FixedPoint(y::T, _) = new(i)
```
timholy: `y->i`? (typo)

src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,198 @@ export
...
+zero{T <: FixedPoint}(::Type{T}) = T(zero(rawtype(T)),0)
+ one{T <: FixedPoint}(::Type{T}) = T(2^nbitsfrac(T)-1,0)
+zero(x::FixedPoint) = zero(typeof(x))
+ one(x::FixedPoint) =  one(typeof(x))
+
+
+# basic operators & arithmetics
+(-){T<:FixedPoint}(x::T) = T(-reinterpret(x), 0)
+(~){T<:FixedPoint}(x::T) = T(~reinterpret(x), 0)
+abs{T<:FixedPoint}(x::T) = T(abs(reinterpret(x.i)),0)
```
timholy: This looks like it will fail.

src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,198 @@ export
...
 realmax{T<: FixedPoint}(::Type{T}) = typemax(T)
+eps{T<:FixedPoint}(::Type{T}) = T(one(rawtype(T)),0)
+eps{T<:FixedPoint}(::T) = eps(T)
+sizeof{T<:FixedPoint}(::Type{T}) = sizeof(rawtype(T))
 
-include("fixed.jl")
-include("ufixed.jl")
-include("deprecations.jl")
+zero{T <: FixedPoint}(::Type{T}) = T(zero(rawtype(T)),0)
+ one{T <: FixedPoint}(::Type{T}) = T(2^nbitsfrac(T)-1,0)
```
timholy: I just noticed that this is probably problematic for `FixedPoint{Int32,32}`. I doubt we can do anything about that, however, without tanking performance elsewhere.

src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,198 @@ export
...
+convert{T,f}(::Type{Fixed{T,f}}, x::Rational) =
+    FixedPoint{T,f}(x.num)/FixedPoint{T,f}(x.den)
+
+# Conversions from FixedPoint
+convert{T,f}(::Type{BigFloat}, x::FixedPoint{T,f}) =
+    convert(BigFloat,x.i>>f) + convert(BigFloat,x.i&(1<<f - 1))/convert(BigFloat,1<<f)
+# convert(::Type{BigFloat}, x::UFixed) = reinterpret(x)*(1/BigFloat(rawone(x)))
+
+convert{TF<:AbstractFloat,T,f}(::Type{TF}, x::FixedPoint{T,f}) =
+    convert(TF,x.i>>f) + convert(TF,x.i&(1<<f - 1))/convert(TF,1<<f)
```
timholy: I think the division will be slower than the former multiplication. Also, try this on `x.i = 0xff`: you get 255/256, which is not quite 1.

timholy:
I assume you're noting it's still failing, seemingly due to the "not quite 1" issue I raised above.

We might need a different implementation of `convert` depending on whether `T<:Signed`.

timholy:
Need any help? I'd love to get this finished, tagged, and then be able to tag new versions for Colors/ColorTypes/ColorVectorSpace/Images etc.

vchuravy:
Any help is appreciated, grad school got a bit in the way :). 

I just pushed the last changes I have around, but I think I am still missing some cases.

So one of the issues is that for signed types `1` is not necessarily part of the representation and maybe using `typemax` and `typemin` is a better approach to convert  `FixedPoint` to `FloatingPoint`.

src/FixedPointNumbers.jl
```diff
@@ -1,4 +1,4 @@
-VERSION >= v"0.4.0-dev+6521" && __precompile__()
+# ERSION >= v"0.4.0-dev+6521" && __precompile__()
```
vchuravy: oops :)

src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,202 @@ export
...
 
+# T => BaseType
+# f => Number of Bytes reserved for fractional part
+immutable FixedPoint{T <: Integer, f} <: Real
+    i::T
+
+    # constructor for manipulating the representation;
+    # selected by passing an extra dummy argument
+    FixedPoint(i::T, _) = new(i)
+    FixedPoint(i::Integer,_) = new(i % T)
```
vchuravy: `i % T` is a new change and maybe it is better to throw a DomainError, at least we should add a test for this

src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,202 @@ export
...
+
+# default provided typealiases
+typealias Fixed16 Fixed{Int32, 16}
+typealias UFixed8  UFixed{UInt8,8}
+typealias UFixed10 UFixed{UInt16,10}
+typealias UFixed12 UFixed{UInt16,12}
+typealias UFixed14 UFixed{UInt16,14}
+typealias UFixed16 UFixed{UInt16,16}
+
 reinterpret(x::FixedPoint) = x.i
```
vchuravy: I was thinking that we maybe should rename this to `get`. `reinterpret` is quite long and one easily writes `x.i` instead

src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,202 @@ export
...
+
++{T<:FixedPoint}(x::T, y::T) = T(reinterpret(x) + reinterpret(y),0)
+-{T<:FixedPoint}(x::T, y::T) = T(reinterpret(x) - reinterpret(y),0)
+
+# with truncation:
+# *{T<:FixedPoint}(x::T, y::T) =
+#         T(Base.widemul(reinterpret(x),reinterpret(y)) >> nbitsfrac(T),0)
+# with rounding up:
+function *{T<:FixedPoint}(x::T, y::T)
+    f = nbitsfrac(T)
```
vchuravy: A was wondering if this is getting optimized out or if we rather should use `@generated` for functions depending on `f` and `T`

src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,202 @@ export
...
+    reinterpret(x)*(1/convert(T, rawone(x)))
+
+convert{T,f}(::Type{Bool}, x::FixedPoint{T,f}) = reinterpret(x) != zero(x)
+
+function convert{TI<:Integer, T,f}(::Type{TI}, x::FixedPoint{T,f})
+    isinteger(x) || throw(InexactError())
+    convert(TI, x.i>>f)
+end
+
+convert{TR<:Rational,T,f}(::Type{TR}, x::FixedPoint{T,f}) =
```
vchuravy: This conversion is probably not correct for `T <: Unsigned`

src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,202 @@ export
...
+# Promote rules
+promote_rule{T <: FixedPoint,TI<:Integer}(::Type{T}, ::Type{TI}) = T
+promote_rule{T <: FixedPoint,TF<:AbstractFloat}(::Type{T}, ::Type{TF}) = TF
+promote_rule{T <: FixedPoint,TR <: Rational}(::Type{T}, ::Type{TR}) = TR
+
+# for T in UF
+#     for Ti in (Int8, UInt8, Int16, UInt16, Int32, UInt32, Int64, UInt64)
+#         Tp = eps(convert(Float32, typemax(Ti))) > eps(T) ? Float64 : Float32
+#         @eval begin
+#             promote_rule(::Type{$T}, ::Type{$Ti}) = $Tp
```
vchuravy: Are these `promote_rule`s needed or do the three on top cover them?

src/FixedPointNumbers.jl
```diff
@@ -42,25 +40,202 @@ export
...
+#         end
+#     end
+# end
 
+# Math functions
+trunc{T<:FixedPoint}(x::T) = T(div(reinterpret(x), rawone(T))*rawone(T),0)
```
vchuravy: This is going to fail for `T <: Signed` where `nbitsfrac(T) == sizeof(T)*8`

vchuravy:
I added some comments, where I still have open questions and not necessary good answers. Essentially every bit of functionality that has a commented out version I was a bit unsure about.

vchuravy:
A fundamental problem I am running into that using `rawone` doesn't make sense for `Fixed`... and it is a bit unintuitive that 0xFF equals 1.0 for `UFixed8 = FixedPoint{UInt8, 8}` where `f = 8` is the number of *fractional* bits, with setting 0xff we are using a fractional bit to signal oneness. 

Using `FixedPoint{UInt8, 7}`  Where the first bit is the integral bit would break the major use of unsigned fixedpoint numbers in colors...

A similar problem arises with signed fixedpoint numbers, where we need one bit to signal sign and so the question is if `FixedPoint{Int8, 8}` is well defined at all.

Maybe my brain is just going off on a tangent and I don't see the forest because of all the trees...

timholy:
You know, this rings a bell and is probably why I didn't unify them when I first wrote `Ufixed`. It has been so long, I had forgotten about these considerations. For colors/images we definitely need `0xff` to correspond to 1, so perhaps these are just two different beasts and should remain separate.

Sound reasonable?

vchuravy:
I am going to think this over, getting rid of a lot of code duplication would be really nice, but I don't really want to break the one major use case of FixedPointNumbers. (Even though UFixed should probably be called normalized numbers ;) )

timholy:
Sounds good. I'll be out of touch for a couple of days.

When I get back, I'll soon go ahead and tag an new release of Images and all related packages, just because it's overdue. If it takes a little longer to figure this out, that's fine---no rush.

vchuravy:
I am still interested in renaming `UFixed` to `UNormed` or `UNorm`. But I don't think anymore that having `Fixed{Unsigned}` is really useful.

timholy:
I could go with such a renaming. (I probably won't do it myself, but happy to see it arrive, and even happier if some helps fix the deprecations it will cause elsewhere :wink:.)

timholy:
If we're thinking about names, some other possible changes to consider:
- rename `FixedPoint` to `AbstractFixedPoint`
- rename `Fixed` to `FixedPoint` (requires the above, and should probably be done in separate minor releases with suitable deprecation warnings and some reasonable time delay...)
- as alternatives to `Normed`, also consider `UnitFixedPoint` (with unit in the sense of a unit vector), `NormalizedFixedPoint`, `FixedPointOne`, `FixedOne`, `FixedZO` (Z=zero, O=one).

PR #42, opened 2016-04-22
Resolve functor deprecation

timholy:


timholy:
Test failure is timeout due to https://github.com/JuliaLang/julia/pull/16012.

PR #43, opened 2016-07-21
Make conversions work for Float16 (and speed tests)

timholy:
Curiously, currently we have this:
```jl

julia> x = UFixed8(0.3)
UFixed8(0.298)

julia> y = convert(Float16, x)
0.29803923f0

julia> typeof(y)
Float32
```
As @vchuravy knows well, that may soon be fixed on julia master (see https://github.com/JuliaLang/julia/pull/17297), but since this package also supports julia 0.4 it seems sensible to fix it here.

I also was reminded about how awfully slow the tests are. Poking at this a bit, I discovered that changing two characters (see 06b6372) speeds up the tests by several fold. Moreover, I suspect there's more improvement possible, because [this loop](https://github.com/JeffBezanson/FixedPointNumbers.jl/blob/f1bafe33e7da24e58edf5c10d49151a0bf4cde96/test/fixed.jl#L37-L43) isn't type-stable, even on julia-0.5 (the type of `fun` is changing on each iteration). However, I confess I'm having trouble understanding the purpose of that loop: as far as I can tell, it only ever runs the tests for `fun == /`. @vchuravy, can you shed some light on this? I'm guessing you meant to run the two lines with `@test` for all 4 functions, skipping it for `/` when `fy == 0`, right?

### DIFF
--- a/src/ufixed.jl
+++ b/src/ufixed.jl
@@ -67,7 +67,10 @@ ufixed16(x) = convert(UFixed16, x)
 
 
 convert(::Type{BigFloat}, x::UFixed) = reinterpret(x)*(1/BigFloat(rawone(x)))
-convert{T<:AbstractFloat}(::Type{T}, x::UFixed) = reinterpret(x)*(1/convert(T, rawone(x)))
+function convert{T<:AbstractFloat}(::Type{T}, x::UFixed)
+    y = reinterpret(x)*(1/convert(T, rawone(x)))
+    convert(T, y)  # needed for types like Float16 which promote arithmetic to Float32
+end
 convert(::Type{Bool}, x::UFixed) = x == zero(x) ? false : true
 convert{T<:Integer}(::Type{T}, x::UFixed) = convert(T, x*(1/one(T)))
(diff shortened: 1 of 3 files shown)


timholy:
OK, I pushed two more commits and have gotten the tests from ~10min to ~20s. @vchuravy, please check to see whether you think these changes are correct.

vchuravy:
Thanks, the for loop was indeed a mistake. `@test` is really slower than `@assert`?

This looks good. lgtm

timholy:
Much of the performance improvement came from other changes, but the `@test`->`@assert` change also had a pretty big impact.

This branch:
```jl
julia> @time include("runtests.jl")
# <output suppressed>
 19.473147 seconds (93.47 M allocations: 1.910 GB, 1.71% gc time)
```
If I drop the last commit that changed `@test` to `@assert`:
```jl
julia> @time include("runtests.jl")
# <output suppressed>
144.961808 seconds (3.16 G allocations: 143.512 GB, 11.94% gc time)
```
I'd say that's enough of an improvement to be worth it. If you disagree, I'm happy to drop the last commit.

vchuravy:
No that looks worth it. Thank you for this.

PR #44, opened 2016-08-12
added support for an arbitrary number of fraction bits

bjarthur:
i need to add four `Image{Gray{UFixed16}}` together without overflowing.  could convert to `Float32`, but thought a `UFixed{UInt32,16}` would be more graceful.  so before realizing that the snazzy new `@generated` was taboo, i refactored `ufixed.jl` to support arbitrary fractional bit depths.  anything goes, like `UFixed{UInt64,3}`, as an additional example.

all of the tests pass in julia 0.5-rc2.

so why are generated functions to be [avoided](https://github.com/JeffBezanson/FixedPointNumbers.jl/pull/35#issuecomment-239555047)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,21 +15,29 @@ used like any other number: they can be added, multiplied, raised to a power,
 etc. In many cases these operations result in conversion to floating-point types.
 
 # Type hierarchy
-This library defines an abstract type `FixedPoint{T <: Integer, f}` as a subtype of `Real`. The parameter `T` is the underlying representation and `f` is the number of fraction bits.
+This library defines an abstract type `FixedPoint{T <: Integer, f}` as a
+subtype of `Real`. The parameter `T` is the underlying representation and `f`
+is the number of fraction bits.
 
-For signed integers, there is a fixed-point type `Fixed{T, f}` and for unsigned integers, there is the `UFixed{T, f}` type.
+For signed integers, there is a fixed-point type `Fixed{T, f}` and for unsigned
+integers, there is the `UFixed{T, f}` type.
 
-These types, built with `f` fraction bits, map the closed interval [0.0,1.0]
-to the span of numbers with `f` bits.
-For example, the `UFixed8` type is represented internally by a `UInt8`, and makes
-`0x00` equivalent to `0.0` and `0xff` to `1.0`.
-The types `UFixed10`, `UFixed12`, `UFixed14`, and `UFixed16` are all based on `UInt16`
-and reach the value `1.0` at 10, 12, 14, and 16 bits, respectively (`0x03ff`, `0x0fff`,
-`0x3fff`, and `0xffff`).
+These types, built with `f` fraction bits, map the closed interval [0.0,1.0] to
+the span of numbers with `f` bits.  For example, the `UFixed8` type (aliased to
+UFixed{UInt8,8}) is represented internally by a `UInt8`, and makes `0x00`
+equivalent to `0.0` and `0xff` to `1.0`.  The type aliases `UFixed10`, `UFixed12`,
+`UFixed14`, and `UFixed16` are all based on `UInt16` and reach the value `1.0`
+at 10, 12, 14, and 16 bits, respectively (`0x03ff`, `0x0fff`, `0x3fff`, and
+`0xffff`).
 
-To construct such a number, use `convert(UFixed12, 1.3)`, `ufixed12(1.3)`, or the literal syntax `0x14ccuf12`.
-The latter syntax means to construct a `UFixed12` (it ends in `uf12`) from the `UInt16` value
-`0x14cc`.
+To construct such a number, use `convert(UFixed12, 1.3)`, `ufixed12(1.3)` (a
+convenience function), `UFixed{UInt16,12}(1.3)`, or the literal syntax
+`0x14ccuf12`.  The latter syntax means to construct a `UFixed12` (it ends in
+`uf12`) from the `UInt16` value `0x14cc`.
+
+More generally, an arbitrary number of bits from any of the standard unsigned
+integer widths can be used for the fractional part.  For example:
+`UFixed{UInt32,16}`, `UFixed{UInt64,3}`, `UFixed{UInt128,7}`.
 
 There currently is no literal syntax for signed `Fixed` numbers. 
 
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -94,4 +94,15 @@ scaledual{T<:FixedPoint}(Tdual::Type, x::Union{T,AbstractArray{T}}) =
 scaledual{Tdual<:Number, T<:FixedPoint}(b::Tdual, x::Union{T,AbstractArray{T}}) =
     convert(Tdual, b/one(T)), reinterpret(rawtype(T), x)
 
+# printing
+function show{T,f}(io::IO, x::FixedPoint{T,f})
+    shorttype = typeof(x)<:UFixed ? "UFixed" : "Fixed"
+    print(io, shorttype, "{", T, ",", f, "}")
+    print(io, "(")
+    showcompact(io, x)
+    print(io, ")")
+end
+const _log2_10 = 3.321928094887362
+showcompact{T,f}(io::IO, x::FixedPoint{T,f}) = show(io, round(convert(Float64,x), ceil(Int,f/_log2_10)))
+
 end # module
--- a/src/ufixed.jl
+++ b/src/ufixed.jl
@@ -108,10 +108,6 @@ trunc{T<:Integer}(::Type{T}, x::UFixed) = convert(T, div(reinterpret(x), rawone(
 round{T<:Integer}(::Type{T}, x::UFixed) = round(T, reinterpret(x)/rawone(x))
 floor{T<:Integer}(::Type{T}, x::UFixed) = trunc(T, x)
  ceil{T<:Integer}(::Type{T}, x::UFixed) =  ceil(T, reinterpret(x)/rawone(x))
-trunc(x::UFixed) = trunc(Int, x)
-round(x::UFixed) = round(Int, x)
-floor(x::UFixed) = floor(Int, x)
- ceil(x::UFixed) =  ceil(Int, x)
 
 isfinite(x::UFixed) = true
 isnan(x::UFixed) = false
--- a/src/ufixed.jl
+++ b/src/ufixed.jl
@@ -19,11 +19,7 @@ typealias UFixed16 UFixed{UInt16,16}
 
 const UF = (UFixed8, UFixed10, UFixed12, UFixed14, UFixed16)
 
-for (uf) in UF
-    T = rawtype(uf)
-    f = nbitsfrac(uf)
-    @eval reinterpret(::Type{UFixed{$T,$f}}, x::$T) = UFixed{$T,$f}(x, 0)
-end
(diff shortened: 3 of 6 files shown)


vchuravy:
A lot of the generated functions are unnecessary and I think we could rework most of the eval-for loops by using methods like `@pure mask{T, f}(::Type{UFixed{T,f}}) = convert(T, 1<<(f-1))`. LLVM constant-folds this calculation and we won't need to resort to the inlining trick.

Will put up an PR for that momentarily. Thank you for your help.

bjarthur:
doh!  it should've occurred to me that generated functions whose body only contains the returned expression could be rewritten as normal functions.  thanks!

i've added more tests, updated the docs, and included a commit, which could be a separate PR if preferred, which fixes the type returned by `round` and `ceil`.

any ideas for benchmarks?

src/ufixed.jl
```diff
@@ -35,14 +31,7 @@ const uf14 = UFixedConstructor{UInt16,14}()
...
 zero{T,f}(::Type{UFixed{T,f}}) = UFixed{T,f}(zero(T),0)
-for uf in UF
-    TT = rawtype(uf)
-    f = nbitsfrac(uf)
-    T = UFixed{TT,f}
-    @eval begin
-        one(::Type{$T}) = $T($(2^f-1),0)
-    end
-end
+one{T<:UFixed}(::Type{T}) = T((2^nbitsfrac(T)-1),0)
```
vchuravy: This will be costly on v0.4. On v0.5 we can constant fold this, but only if we outline the `2^nbitsfrac(T)` into a pure function `Base.@pure one_pattern(f)=(2^f-1)`, but this won't work on v0.4.

One solution would be for now to use the current way on v0.4 with your method as a fallback, and on v0.5 and higher use pure functions on the type-parameter to get the same effect.

src/ufixed.jl
```diff
@@ -95,26 +84,28 @@ abs(x::UFixed) = x
...
-for T in UF
-    f = nbitsfrac(T)
-    R = rawtype(T)
-    roundmask = convert(R, 1<<(f-1))
-    k = 8*sizeof(R)-f
-    ceilmask  = (typemax(R)<<k)>>k
-    @eval begin
-        round(x::$T) = (y = trunc(x); return convert(rawtype($T), reinterpret(x)-reinterpret(y))&$roundmask>0 ? $T(y+one($T)) : y)
-         ceil(x::$T) = (y = trunc(x); return convert(rawtype($T), reinterpret(x)-reinterpret(y))&$ceilmask >0 ? $T(y+one($T)) : y)
+@generated function round{T,f}(x::UFixed{T,f})
```
vchuravy: These generated functions are unnecessary on v0.5 if you out-line the calculations into pure functions.

src/ufixed.jl
```diff
@@ -152,25 +143,8 @@ function decompose(x::UFixed)
...
-    print(io, "(")
-    showcompact(io, x)
-    print(io, ")")
-end
-showcompact{T,f}(io::IO, x::UFixed{T,f}) = show(io, round(convert(Float64,x), ceil(Int,f/_log2_10)))
+promote_rule{T<:UFixed}(::Type{T}, ::Type{Float32}) = Float32
+promote_rule{T<:UFixed}(::Type{T}, ::Type{Float64}) = Float64
+promote_rule{T<:UFixed, R<:Rational}(::Type{T}, ::Type{R}) = R
+promote_rule{T<:UFixed, Ti<:Union{Signed,Unsigned}}(::Type{T}, ::Type{Ti}) =
+        eps(convert(Float32, typemax(Ti))) > eps(T) ? Float64 : Float32
```
vchuravy: Is `eps(convert(Float32, typemax(Ti)))` getting constant folded?

test/ufixed.jl
```diff
@@ -113,7 +125,7 @@ function testtrunc{T}(inc::T)
     end
 end
 
-for T in FixedPointNumbers.UF
+for T in (FixedPointNumbers.UF..., UF2...)
```
vchuravy: We probably shouldn't test UF2 here since `testtrunc` tests every value under the sun and increases the testing times massively.

test/ufixed.jl
```diff
@@ -122,7 +134,7 @@ x = 0xaauf8
 iob = IOBuffer()
 show(iob, x)
 str = takebuf_string(iob)
-@test startswith(str, "UFixed8(")
+@test startswith(str, "FixedPointNumbers.UFixed{UInt8,8}(")
```
vchuravy: Can we turn this into `UFixed{UInt8,8}`? Since this gets used in Images.jl and the type names are already quite long.

vchuravy:
Thank you for tackling this. The added generality will be valuable, but we have to make sure that this comes at *no* extra cost, because FixedPointNumbers is used in performance sensitive places.

Would you mind also adding a few benchmarks with `BenchmarkTools`?

src/ufixed.jl
bjarthur: good catch.  thanks.  it was 4x slower.   fixed for 0.4 and 0.5 with a generated function.  that seems a simpler solution to me than having to switch on the version.

src/ufixed.jl
bjarthur: i don't see what's wrong with generated functions so long as they don't have side effects and don't reference global variables.  the code as is works with 0.4 and 0.5.

src/ufixed.jl
bjarthur: probably not.  good catch.  thanks again.  have fixed with another generated function.

test/ufixed.jl
bjarthur: note the `min(1e6, ...`

bjarthur:
thanks for pointing out BenchmarkTools to me.  i used it to fix a few things you pointed out.  is it customary to commit benchmarks?  if so, where?

src/ufixed.jl
```diff
@@ -98,26 +92,28 @@ abs(x::UFixed) = x
...
+        return convert(T, reinterpret(x)-reinterpret(y)) & ($mask)>0 ?
+                UFixed{T,f}(y+one(UFixed{T,f})) : y
     end
 end
 
 trunc{T<:Integer}(::Type{T}, x::UFixed) = convert(T, div(reinterpret(x), rawone(x)))
 round{T<:Integer}(::Type{T}, x::UFixed) = round(T, reinterpret(x)/rawone(x))
 floor{T<:Integer}(::Type{T}, x::UFixed) = trunc(T, x)
  ceil{T<:Integer}(::Type{T}, x::UFixed) =  ceil(T, reinterpret(x)/rawone(x))
-trunc(x::UFixed) = trunc(Int, x)
```
vchuravy: Any reason for these to be removed?

src/ufixed.jl
vchuravy: I am not the expert her so maybe @JeffBezanson can chime in. But my understanding is that generated functions are best to be avoided and are indeed unnecessary for this on `v0.5`.

bjarthur: what about 0.4?

src/ufixed.jl
bjarthur: this is part of the second commit.  removing them fixes the bug that round(UFixed) returns an Int, not a UFixed.

test/ufixed.jl
timholy: I say truncate it. As Jeff says, printing of *types* should include all the gory details. Printing of *values* can take shortcuts.

timholy:
This looks quite good to me. I do not personally have a problem with the generated functions. Some could become `@pure` on Julia 0.5, of course.

bjarthur:
`show` now omits `FixedPointNumbers` for brevity, as requested.

anything else?

src/FixedPointNumbers.jl
```diff
@@ -94,4 +94,17 @@ scaledual{T<:FixedPoint}(Tdual::Type, x::Union{T,AbstractArray{T}}) =
 scaledual{Tdual<:Number, T<:FixedPoint}(b::Tdual, x::Union{T,AbstractArray{T}}) =
     convert(Tdual, b/one(T)), reinterpret(rawtype(T), x)
 
+# printing
+@generated function show{T,f}(io::IO, x::FixedPoint{T,f})
```
vchuravy: Why is this a `@generated` function? I would think it is unnecessary here ;)

vchuravy:
Looks good to me.

src/FixedPointNumbers.jl
timholy: Right, this should not be `@generated`, it's not performance-critical and in any case being `@generated` won't help here.

bjarthur:
is this what you had in mind instead of a generated function?

README.md
```diff
@@ -15,21 +15,29 @@ used like any other number: they can be added, multiplied, raised to a power,
...
+These types, built with `f` fraction bits, map the closed interval [0.0,1.0] to
+the span of numbers with `f` bits.  For example, the `UFixed8` type (aliased to
+UFixed{UInt8,8}) is represented internally by a `UInt8`, and makes `0x00`
+equivalent to `0.0` and `0xff` to `1.0`.  The type aliases `UFixed10`, `UFixed12`,
+`UFixed14`, and `UFixed16` are all based on `UInt16` and reach the value `1.0`
+at 10, 12, 14, and 16 bits, respectively (`0x03ff`, `0x0fff`, `0x3fff`, and
+`0xffff`).
+
+To construct such a number, use `convert(UFixed12, 1.3)`, `ufixed12(1.3)` (a
+convenience function), `UFixed{UInt16,12}`, or the literal syntax `0x14ccuf12`.
```
timholy: `UFixed{UInt16,12}` is a type, not a value.

src/ufixed.jl
timholy: Hmm, I just checked, and in my hands (on julia 0.5) the non-generated and generated versions produce identical `@code_llvm`. Do you see something different on julia 0.4?

src/ufixed.jl
```diff
@@ -155,25 +149,10 @@ function decompose(x::UFixed)
...
-function show{T,f}(io::IO, x::UFixed{T,f})
-    print(io, "UFixed", f)
-    print(io, "(")
-    showcompact(io, x)
-    print(io, ")")
+promote_rule{T<:UFixed}(::Type{T}, ::Type{Float32}) = Float32
+promote_rule{T<:UFixed}(::Type{T}, ::Type{Float64}) = Float64
+promote_rule{T<:UFixed, R<:Rational}(::Type{T}, ::Type{R}) = R
+@generated function promote_rule{T<:UFixed, Ti<:Union{Signed,Unsigned}}(::Type{T}, ::Type{Ti})
+    Tp = eps(convert(Float32, typemax(Ti))) > eps(T) ? Float64 : Float32
```
timholy: This one definitely needs to be `@generated`.

timholy:
Sorry to keep pestering you with review comments. I was about to hit merge and thought I should give it another once-over. I must be less tired this time.

bjarthur:
dang.  good catches.  two less evil generated functions now.  0.4 behaves the same.

timholy:
Thanks!

BTW, to answer your question above about why they are evil:
- They are much slower to JIT compile, so add to latency-to-first-result when using them
- They prevent static compilation (see http://juliacomputing.com/blog/2016/02/09/static-julia.html)
- Jameson hates them :smile:

All of these actually come down to the same thing: the compiler can't easily reason about what a generated function will do, because they can do arbitrary things.

So really they're only partly evil, but it's fair to say they are to be avoided unless there is no really good alternative.

PR #46, opened 2016-09-05
Fix vectorize function depwarn on 0.6

yuyichao:
Requires https://github.com/JuliaLang/Compat.jl/pull/280 and a new `Compat` tag.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
 julia 0.4
-Compat 0.7.14
+Compat 0.9.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
     - osx
 julia:
     - 0.4
+    - 0.5
     - nightly
 notifications:
   email: false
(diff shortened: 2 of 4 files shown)


REQUIRE
```diff
@@ -1,2 +1,2 @@
 julia 0.4
-Compat 0.7.14
+Compat 0.8.9
```
tkelman: this won't be right

yuyichao: 0.8.9 wasn't there when I made the change 8 hours ago.

tkelman: it's not going to be 0.8.10 either

yuyichao: Well, I don't really care what it is going to be right now. So if I can't put a place holder like `0.8.8+` and I can't just put what is currently the next version should I just leave the version requirement unchanged?

tkelman: no, better to hold off on opening the pr if it shouldn't be merged yet

yuyichao:
Compat.jl PR merged and tagged. Version requirements updated and CI passed. Should be ready to go.

Also updated travis to also run on 0.5.

PR #47, opened 2016-09-11
Fix div and friends

timholy:
Gracious me, coding by copy/paste does have its downsides.

This one was my fault from back in 2014, I believe. Amazing it's not been noticed before.

### DIFF
--- a/src/ufixed.jl
+++ b/src/ufixed.jl
@@ -116,7 +116,12 @@ isinf(x::UFixed) = false
 bswap{f}(x::UFixed{UInt8,f}) = x
 bswap(x::UFixed)  = typeof(x)(bswap(reinterpret(x)),0)
 
-for f in (:div, :fld, :rem, :mod, :mod1, :rem1, :fld1, :min, :max)
+for f in (:div, :fld, :fld1)
+    @eval begin
+        $f{T<:UFixed}(x::T, y::T) = $f(reinterpret(x),reinterpret(y))
+    end
+end
+for f in (:rem, :mod, :mod1, :rem1, :min, :max)
     @eval begin
         $f{T<:UFixed}(x::T, y::T) = T($f(reinterpret(x),reinterpret(y)),0)
     end
--- a/test/ufixed.jl
+++ b/test/ufixed.jl
@@ -133,6 +133,18 @@ for T in (FixedPointNumbers.UF..., UF2...)
     testtrunc(eps(T))
 end
 
+@test div(0x10uf8, 0x02uf8) == fld(0x10uf8, 0x02uf8) == 8
+@test div(0x0fuf8, 0x02uf8) == fld(0x0fuf8, 0x02uf8) == 7
+@test Base.fld1(0x10uf8, 0x02uf8) == 8
+@test Base.fld1(0x0fuf8, 0x02uf8) == 8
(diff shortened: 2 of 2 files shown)


vchuravy:
I am not on a computer right now so I couldn't just test this :)

Was the problem here the return type of the `div` operator? It seems odd to me that the result would be an integer (in all cases).

timholy:
This is what happens on master:
```jl
INFO: Recompiling stale cache file /home/tim/.julia/lib/v0.5/FixedPointNumbers.ji for module FixedPointNumbers.

julia> x = 16uf8
UFixed{UInt8,8}(0.063)

julia> y = 2uf8
UFixed{UInt8,8}(0.008)

julia> div(x, y)
UFixed{UInt8,8}(0.031)

julia> div(Float32(x), Float32(y))
8.0f0

julia> r = 1uf8:1uf8:10uf8
UFixed{UInt8,8}(0.004):UFixed{UInt8,8}(0.004):UFixed{UInt8,8}(0.039)

julia> length(r)
------ InexactError -------------------- Stacktrace (most recent call last)

 [1] — length(::StepRange{FixedPointNumbers.UFixed{UInt8,8},FixedPointNumbers.UFixed{UInt8,8}}) at range.jl:364

 [2] — unsafe_length(::StepRange{FixedPointNumbers.UFixed{UInt8,8},FixedPointNumbers.UFixed{UInt8,8}}) at range.jl:361

InexactError()
```
None of those problems happen with this branch.

vchuravy:
Hm I see. I am still trying to wrap my head around this. Isn't the problem
that the result of the division is not representable as a UF8? I am still
wondering how falling back to integer division solves this, wouldn't it
make more sense to fall back to floating point division?

Can I also note that the printing of `2uf8` is broken on master? Or the
conversion to Float32.

src/ufixed.jl
```diff
@@ -116,7 +116,12 @@ isinf(x::UFixed) = false
 bswap{f}(x::UFixed{UInt8,f}) = x
 bswap(x::UFixed)  = typeof(x)(bswap(reinterpret(x)),0)
 
-for f in (:div, :fld, :rem, :mod, :mod1, :rem1, :fld1, :min, :max)
+for f in (:div, :fld, :fld1)
+    @eval begin
+        $f{T<:UFixed}(x::T, y::T) = $f(reinterpret(x),reinterpret(y))
```
vchuravy: This is what confuses me. The result type here is now an unsigned integer.

timholy:
Yes, it's not representable, but in general you expect it not to be. The definition of `div` is just "Computes x/y, truncated to an integer." So the result should be 8, not 0.031.

I'm not sure I follow what's broken about the printing/conversion to Float32:
```jl
julia> x = 2uf8
UFixed{UInt8,8}(0.008)

julia> Float32(x)
0.007843138f0
```
(except I'm bummed about how verbose the printing is, I thought we'd fixed that)

src/ufixed.jl
timholy: Are you arguing we should throw an `InexactError`?

vchuravy:
Ahhh, I am too tired. You are of course right.
- `div` should return integers.
- and I compared `Float32(2uf8)` to the wrong number

Thanks for explaining. LGTM  from my side.

Issue #37, opened 2015-12-08
`convert` sometimes throws InexactError, usually rounds


ssfrr:
`convert(Fixed{Int8, 7}, x)` seems to happily round to the nearest representable value for most values of `x`, but throws an `InexactError` for `x=0.999`. AFAICT it throws the error when the value would get rounded up to `1.0`, but `1.0` can't be represented. Is this expected behavior? I would expect that in this case it would act the same as `convert(Fixed{Int8, 7}, 1.0)`, which overflows to `-1.0`.

```
julia> convert(Fixed{Int8, 7}, 0.8)
FixedPointNumbers.Fixed{Int8,7}(0.797)

julia> convert(Fixed{Int8, 7}, 0.9)
FixedPointNumbers.Fixed{Int8,7}(0.898)

julia> convert(Fixed{Int8, 7}, 0.999)
ERROR: InexactError()
 in trunc at float.jl:357
 in convert at /Users/srussell/.julia/v0.4/FixedPointNumbers/src/fixed.jl:34

julia> convert(Fixed{Int8, 7}, 1.0)
FixedPointNumbers.Fixed{Int8,7}(-1.0)
```

should `convert` accept values that can't be exactly represented? My understanding was that normally `convert(T, x)` is supposed to preserve information and thrown an error if the value can't be represented, and `round(T, x)` is supposed to be used to round to the nearest representable number, but I don't think that's explicitly said, so I could be wrong there.

timholy:
Yeah, this seems a bit broken. The corresponding base-julia behavior:
```jl
julia> convert(UInt8, 256)
ERROR: InexactError()
 in convert at int.jl:175

julia> 256 % UInt8
0x00
```

Issue #21, opened 2015-05-14
Ufixed8(0xff) isn't Ufixed8(1.0)


yakir12:
```Julia 
julia> Ufixed8(0xff)
Ufixed8(0.004)
```
Isn't it supposed to be Ufixed8(1.0)?
```Julia 
Julia Version 0.4.0-dev+4791
Commit a9b0135 (2015-05-12 15:32 UTC)
Platform Info:
  System: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-3520M CPU @ 2.90GHz
  WORD_SIZE: 64
  BLAS: libopenblas (NO_LAPACK NO_LAPACKE DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: liblapack.so.3
  LIBM: libopenlibm
  LLVM: libLLVM-3.3
```
`julia runtests.jl` completed without any errors though..?

timholy:
You're being tripped up by the difference between conversion and reinterpretation:
```jl
julia> Ufixed8(0xff)
Ufixed8(0.004)

julia> reinterpret(Ufixed8, 0xff)
Ufixed8(1.0)

julia> 0xffuf8
Ufixed8(1.0)
```
On julia 0.4, we should probably have `Ufixed8(0xff)` throw an error, since 255 is not representable as a Ufixed8 (it's still using the old julia 0.3 "unsafe conversion" convention).

yakir12:
Ah, I see. I'll try and watch out for that. That settles that then.
Thanks!

Issue #15, opened 2014-10-08
Test fails on 32-bit Windows 7


gyk:
Running test in Julia v0.3:

``` Julia
julia> Pkg.test("FixedPointNumbers")
INFO: Testing FixedPointNumbers
ERROR: test failed: one(T) == 1
while loading \.julia\v0.3\FixedPointNumbers\test\ufixed.jl, in expression starting on line 20
while loading \.julia\v0.3\FixedPointNumbers\test\runtests.jl, in expression starting on line 2
...
ERROR: FixedPointNumbers had test errors

julia> one(Ufixed8)
Ufixed8(1.0)

julia> ans == 1
false
```

However, the following code works as expected:

``` Julia
julia> f() = Ufixed8(1.0) == 1
f (generic function with 2 methods)

julia> f()
true

julia> apply(==, promote(one(Ufixed8), 1))
true
```

vchuravy:
this should be fixed now

PR #48, opened 2016-09-14
Use widening to reliably throw InexactErrors

timholy:
It turns out we still have the behavior described in #21 and #37. Hopefully this PR fixes those issues.

Since this is a change in behavior, I suspect the next tag should be a minor increment (even though @vchuravy and @bjarthur have contributed much more "fundamental" changes, at least their changes were non-breaking). Consequently, now might be a good time to consider other disruptive changes. I therefore also deprecated the lowercase conversion functions `ufixed8`. I'll probably also work on trying to come up with a more compact printing syntax before we tag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
 # FixedPointNumbers
 
-This library exports fixed-point number types.
-A [fixed-point number][wikipedia] represents a fractional, or non-integral, number.
-In contrast with the more widely known floating-point numbers, fixed-point
-numbers have a fixed number of digits (bits) after the decimal (radix) point.
-They are effectively integers scaled by a constant factor.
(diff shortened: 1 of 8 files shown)


timholy:
@JeffBezanson, it looks like you're the only one with "admin" privileges. Can you please turn on the Appveyor webook? I'm hoping to find out whether #15 is still an issue.

bjarthur:
need to take `ufixed12` out of the README

timholy:
Done. I took the opportunity to expand the amount of explanation about the meaning of these numbers. Had we done this earlier, it might have saved @vchuravy a certain amount of suffering (#32).

vchuravy:
These changes look great and the additional explanations helps a lot.

PR #53, opened 2016-09-23
Implement promotion among UFixed, make convert errors more informative

timholy:
On the master branch, we have the following bug:
```jl
julia> UFixed8(0.5) < UFixed16(0.8)
ERROR: no promotion exists for FixedPointNumbers.UFixed{UInt8,8} and FixedPointNumbers.UFixed{UInt16,16}
 in promote(::FixedPointNumbers.UFixed{UInt8,8}, ::FixedPointNumbers.UFixed{UInt16,16}) at ./promotion.jl:153
 in <(::FixedPointNumbers.UFixed{UInt8,8}, ::FixedPointNumbers.UFixed{UInt16,16}) at ./promotion.jl:204
```

We also have this uninformative message:
```jl
julia> convert(UFixed8, 1.2)
ERROR: InexactError()
 in trunc(::Type{UInt8}, ::Float64) at ./float.jl:458
 in _convert at /home/tim/.julia/v0.5/FixedPointNumbers/src/ufixed.jl:47 [inlined]
 in convert(::Type{FixedPointNumbers.UFixed{UInt8,8}}, ::Float64) at /home/tim/.julia/v0.5/FixedPointNumbers/src/ufixed.jl:46
```

This PR fixes the bug by implementing `promote_type` among `UFixed` types. It also produces the more verbose error message:
```jl
julia> convert(UFixed8, 1.2)
ERROR: ArgumentError: FixedPointNumbers.UFixed{UInt8,8} is an 8-bit type representing 256 values from 0.0 to 1.0; cannot represent 1.2
 in throw_converterror(::Type{FixedPointNumbers.UFixed{UInt8,8}}, ::Float64) at /home/tim/.julia/v0.5/FixedPointNumbers/src/FixedPointNumbers.jl:118
 in _convert(::Type{FixedPointNumbers.UFixed{UInt8,8}}, ::Type{UInt8}, ::Float64) at /home/tim/.julia/v0.5/FixedPointNumbers/src/ufixed.jl:53
 in convert(::Type{FixedPointNumbers.UFixed{UInt8,8}}, ::Float64) at /home/tim/.julia/v0.5/FixedPointNumbers/src/ufixed.jl:50
```

Assuming people like this, I think it should be merged before #51, and we can tag a new version before making a more incompatible change.

### DIFF
--- a/src/ufixed.jl
+++ b/src/ufixed.jl
@@ -151,3 +151,15 @@ promote_rule{T<:UFixed, R<:Rational}(::Type{T}, ::Type{R}) = R
     Tp = eps(convert(Float32, typemax(Ti))) > eps(T) ? Float64 : Float32
     :( $Tp )
 end
+@generated function promote_rule{T1,T2,f1,f2}(::Type{UFixed{T1,f1}}, ::Type{UFixed{T2,f2}})
+    f = max(f1, f2)  # ensure we have enough precision
+    T = promote_type(T1, T2)
+    # make sure we have enough integer bits
+    i1, i2 = 8*sizeof(T1)-f1, 8*sizeof(T2)-f2  # number of integer bits for each
+    i = 8*sizeof(T)-f
+    while i < max(i1, i2)
+        T = widen1(T)
+        i = 8*sizeof(T)-f
+    end
+    :(UFixed{$T,$f})
+end
--- a/src/ufixed.jl
+++ b/src/ufixed.jl
@@ -9,7 +9,9 @@ immutable UFixed{T<:Unsigned,f} <: FixedPoint{T,f}
 end
 
   rawtype{T,f}(::Type{UFixed{T,f}}) = T
+  rawtype(x::Number) = rawtype(typeof(x))
 nbitsfrac{T,f}(::Type{UFixed{T,f}}) = f
+nbitsfrac(x::Number) = nbitsfract(typeof(x))
 
 typealias UFixed8  UFixed{UInt8,8}
 typealias UFixed10 UFixed{UInt16,10}
@@ -31,9 +33,8 @@ const uf14 = UFixedConstructor{UInt16,14}()
 const uf16 = UFixedConstructor{UInt16,16}()
 
 zero{T,f}(::Type{UFixed{T,f}}) = UFixed{T,f}(zero(T),0)
-@generated function one{T<:UFixed}(::Type{T})
-    f = 2^nbitsfrac(T)-1
-    :( T($f,0) )
+function one{T<:UFixed}(::Type{T})
+    T(typemax(rawtype(T)) >> (8*sizeof(T)-nbitsfrac(T)), 0)
 end
 zero(x::UFixed) = zero(typeof(x))
  one(x::UFixed) =  one(typeof(x))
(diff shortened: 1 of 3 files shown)


timholy:
Closing and reopening to see if AppVeyor triggers.

src/ufixed.jl
```diff
@@ -41,15 +41,27 @@ rawone(v) = reinterpret(one(v))
...
 convert{U<:UFixed}(::Type{U}, x::Real) = _convert(U, rawtype(U), x)
-_convert{U<:UFixed,T}(::Type{U}, ::Type{T}, x)       = U(round(T, widen1(rawone(U))*x), 0)
-_convert{U<:UFixed  }(::Type{U}, ::Type{UInt128}, x) = U(round(UInt128, rawone(U)*x), 0)
+function _convert{U<:UFixed,T}(::Type{U}, ::Type{T}, x)
+    y = round(widen1(rawone(U))*x)
+    (0 <= y) & (y <= typemax(T)) || throw_converterror(U, x)
+    U(_unsafe_trunc(T, y), 0)
+end
+function _convert{U<:UFixed}(::Type{U}, ::Type{UInt128}, x)
+    y = round(rawone(U)*x)   # for UInt128, we can't widen
```
timholy: I thought about that. The flip side is that if one is counting on `widen1` to actually protect you, then you'll be mistaken.

But perhaps that's at too low of a level to matter, and that the only issue that matters is whether `convert` successfully catches the problem. Which refusing to defining `widen1` doesn't actually help with. I will change it.

timholy: OK, we had a number of problems with high-precision UFixed types.

src/ufixed.jl
```diff
@@ -35,25 +37,38 @@ zero{T,f}(::Type{UFixed{T,f}}) = UFixed{T,f}(zero(T),0)
...
 zero(x::UFixed) = zero(typeof(x))
  one(x::UFixed) =  one(typeof(x))
 rawone(v) = reinterpret(one(v))
 
 # Conversions
 convert{T<:UFixed}(::Type{T}, x::T) = x
-convert{T1<:UFixed}(::Type{T1}, x::UFixed) = reinterpret(T1, round(rawtype(T1), (rawone(T1)/rawone(x))*reinterpret(x)))
+function convert{T<:UFixed}(::Type{T}, x::UFixed)
+    y = round((rawone(T)/rawone(x))*reinterpret(x))
+    (0 <= y) & (y <= typemax(rawtype(T))) || throw_converterror(T, x)
```
bjarthur: why not just say `(0 <= y <= typemax(rawtype(T))) || ...` ?

timholy: Performance. Currently `a <= x <= b` gets lowered to `(a <= x) && (x <= b)` rather than `(a <= x) & (x <= b)`, which means there are two branches rather than one.

This is something that should probably be changed in Julia, but neither I nor anyone else has gotten to it yet.

src/ufixed.jl
```diff
@@ -31,29 +33,40 @@ const uf14 = UFixedConstructor{UInt16,14}()
 const uf16 = UFixedConstructor{UInt16,16}()
 
 zero{T,f}(::Type{UFixed{T,f}}) = UFixed{T,f}(zero(T),0)
-@generated function one{T<:UFixed}(::Type{T})
-    f = 2^nbitsfrac(T)-1
-    :( T($f,0) )
+function one{T<:UFixed}(::Type{T})
+    T(typemax(rawtype(T)) >> (8*sizeof(T)-nbitsfrac(T)), 0)
```
bjarthur: the generated version might be faster.  see https://github.com/JuliaMath/FixedPointNumbers.jl/pull/44#r74753925.  or does the use of `>>` instead of `^` close the gap?

timholy: LLVM inlines this new version to a constant:
```jl
julia> @code_llvm one(UFixed{UInt64, 51})

define %UFixed @julia_one_70810(%jl_value_t*) #0 {
top:
  ret %UFixed { i64 2251799813685247 }
}
```
The bigger issue was that for `UInt128` (or even for `UInt64` on 32-bit machines), `2^f` was returning 0 for `f` bigger than the `WORD_SIZE`. This version is guaranteed to be safe no matter what kind of machine or what kind of precision we're using.

PR #54, opened 2016-09-24
Turn on coverage-testing for Julia 0.6, add README badges

timholy:
We restrict to 0.6 because earlier versions of Julia could not obtain accurate coverage info without turning off inlining, and for this package inlining makes a big difference in the running time of the tests.


timholy:
Coverage testing -> add more tests to improve coverage -> discover yet more bugs :frowning:

PR #59, opened 2016-10-08
Centralize floating-point conversion and implement isapprox

timholy:
This tries to be a little more consistent about *which* floating-point type (`Float32` or `Float64`) we use for conversions. As a consequence, this has the possibility of being breaking :frowning_face:, but I personally think it's worth it. I'll give others a chance to chime in before merging.

Also implemented `isapprox` so `x ≈ y` works nicely in tests (by default, allowing a 1-eps difference).

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -4,7 +4,7 @@ module FixedPointNumbers
 
 using Base: reducedim_initarray
 
-import Base: ==, <, <=, -, +, *, /, ~,
+import Base: ==, <, <=, -, +, *, /, ~, isapprox,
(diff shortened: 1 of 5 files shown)


vchuravy:
This looks good thank you for tackling this.

tkelman:
Guess nobody has complained for almost a year, but this supported 0.4 and has been giving ambiguity warnings there.

timholy:
Is there anything we need to do? The current FPN requires Julia 0.5.

tkelman:
Right but the last 0.4-supporting versions was after this. If it's a one liner to fix, might be worth backporting to a branch off that last 0.4-supporting version and doing one last tag in that old minor series.

PR #61, opened 2016-11-06
implement rand()

bjarthur:
this should set the stage for a subsequent effort to make `rand(Gray{UFixed{UInt16,14}}, 3)` work.


timholy:
LGTM, but notice the test failures.

bjarthur:
i've simplified the test in this most recent force push.

also, please tell me what you think of the hack around JuliaLang/julia#10140

bjarthur:
i removed the hack, and benchmarked.  much faster than the original PR now.  i think it's ready to go.  will wait a few days for comments before merging.

PR #63, opened 2017-01-28
Rename UFixed to Normed

vchuravy:
This rename reflects the nature of `UFixed` it is not just a unsigned
fixpoint number type (for that you can use `Fixed{T<:Unsigned}`), but
rather a number type that is normalised to a specific `one`.


timholy:
Really glad you did this. I was worried you weren't going to be smart about this; this way, *I* have to deal with all the deprecation warnings in ColorTypes and so forth, rather than you. Smart man.

vchuravy:
Well, what can I say.  It seemed easy enough to do it now. ;) Ping me if
you need help with something and I would be interested in seeing if we can
get good performance out of FixedpointNumbers with CUDAnative.

PR #65, opened 2017-01-28
Drop Julia 0.4 support

tkelman:
avoid introducing new deprecations on a version of Julia that
many packages are not introducing new releases on

ref https://github.com/JuliaLang/METADATA.jl/pull/7724

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.4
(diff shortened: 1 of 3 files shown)


vchuravy:
Fine with me, but I think @timholy has still a large code-base that is v0.4 only?

timholy:
No, we've switched (this week) to julia 0.5. So I'm all in favor of this.

PR #74, opened 2017-02-11
Fix an ambiguity on julia 0.6

timholy:
Sometimes, it's the simplest things that fail. Detected in a test failure for ImageFiltering, https://travis-ci.org/JuliaImages/ImageFiltering.jl/jobs/200619941.

### DIFF
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -36,7 +36,7 @@ zero(x::Normed) = zero(typeof(x))
 rawone(v) = reinterpret(one(v))
 
 # Conversions
-convert{T<:Normed}(::Type{T}, x::T) = x
+convert{T<:Unsigned,f}(::Type{Normed{T,f}}, x::Normed{T,f}) = x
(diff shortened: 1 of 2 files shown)


vchuravy:
We have the ambiguity tests currently turned off for v0.6 we should probably turn them on again, (there might have been a Base issue though)

timholy:
Agreed. Currently the problem is that Base reports a lot of ambiguities right now. Some of them are fictitious (https://github.com/JuliaLang/julia/issues/20246, https://github.com/JuliaLang/julia/issues/20567), others may be real.

I've sometimes used `setdiff` to "subtract off" ambiguities in Base from those in other modules. But when you're extending a Base method, that won't help.

timholy:
Should have mentioned, the second of those bug reports was motivated by trying to turn on the ambiguity tests here. :smile:

timholy:
We'll want to tag again soon, but ideally we'd incorporate #73.

PR #80, opened 2017-03-01
Force inlining on convert and rem

timholy:
This leads to a ~30% performance improvement in certain operations. It has a noticeable effect on visualizations.

This doesn't make conversions "perfect," though:
```julia
julia> using FixedPointNumbers
                                                                                   
julia> foo(x) = convert(N0f8, convert(Float32, x))   # in an ideal world, LLVM figures out this is a no-op
foo (generic function with 1 method)                                               
                                                                                   
julia> x = N0f8(0.8)
0.8N0f8                                                                            
                                                                                   
julia> @code_native foo(x)
        .text                                                                      
Filename: REPL[2]                                                                  
        pushq   %rbp                                                               
        movq    %rsp, %rbp                                                         
Source line: 66                                                                    
        movzbl  (%rdi), %eax                                                       
        vcvtsi2ssl      %eax, %xmm0, %xmm0                                         
        movabsq $139739572855192, %rax  # imm = 0x7F17A799E198                     
        vmulss  (%rax), %xmm0, %xmm0                                               
        movabsq $139739572855196, %rax  # imm = 0x7F17A799E19C                     
Source line: 47                                                                    
        vmovss  (%rax), %xmm2           # xmm2 = mem[0],zero,zero,zero             
        vmulss  %xmm2, %xmm0, %xmm1                                                
        vroundss        $4, %xmm1, %xmm0, %xmm1                                    
Source line: 48                                                                    
        vucomiss        %xmm1, %xmm2                                               
        jb      L71
        vxorps  %xmm2, %xmm2, %xmm2
        vucomiss        %xmm2, %xmm1
        jb      L71
Source line: 1
        vcvttss2si      %xmm1, %eax
        popq    %rbp
        retq
Source line: 48
L71:
        movabsq $throw_converterror, %rax
        movabsq $139739908406960, %rdi  # imm = 0x7F17BB99FEB0
        callq   *%rax
        nopl    (%rax)
```

PR #83, opened 2017-04-20
Revert "Merge pull request #80 from JuliaMath/teh/perf"

### LINKED ISSUE #82: Array iteration is slow
Here's an example:

```julia
julia> f = x->begin 
              i=0;
              for k in eachindex(x)
                   i+=x[k]
              end
       end
(::#21) (generic function with 1 method)

julia> d = rand(N2f14, 100,100,100,100);

julia> @time f(d);
 11.956324 seconds (700.00 M allocations: 10.431 GB, 10.25% gc time)

julia> using Images
julia> @time f(rawview(d));
  0.128216 seconds (1.73 k allocations: 81.318 KB)
```

I'm getting by for now with using `Images.rawview` when iterating through `N2f14` arrays, but I expected that wouldn't be necessary.

timholy:
This gets rid of the "optimizations" in #80. Fixes #82.

### DIFF
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -33,43 +33,42 @@ rawone(v) = reinterpret(one(v))
 
 # Conversions
 convert{T<:Unsigned,f}(::Type{Normed{T,f}}, x::Normed{T,f}) = x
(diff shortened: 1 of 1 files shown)


timholy:
CC @Cody-G

PR #100, opened 2018-01-16
Update reduction machinery for 0.7

simonbyrne:
The promotion machinery for reduction in sum/prod was changed in https://github.com/JuliaLang/julia/pull/25051. This updates the use.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,6 +1,8 @@
 using FixedPointNumbers, Base.Test
 
-@test isempty(detect_ambiguities(FixedPointNumbers, Base, Core))
+if VERSION < v"0.7.0-"
+    @test isempty(detect_ambiguities(FixedPointNumbers, Base, Core))
+end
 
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -1,6 +1,8 @@
 using FixedPointNumbers, Base.Test
 
-@test isempty(detect_ambiguities(FixedPointNumbers, Base, Core))
+if VERSION < v"0.7.0-"
+	@test isempty(detect_ambiguities(FixedPointNumbers, Base, Core))
```
ararslan: Tab here

andreasnoack:
I've defined a few methods to resolve some ambiguities and then I've marked a few printing methods as broken to get this one ready. There are still some ambiguities as reported in https://github.com/JuliaMath/FixedPointNumbers.jl/issues/101 but I think we should merge this as is and then resolve the ambiguities later since this is blocking work other packages.

timholy:
Thanks folks!

PR #107, opened 2018-06-05
WIP: Start on 0.7 compatibility

Keno:
This package is fairly stable, so I think the best way to go about
this is to bump master to 0.7. If we intend to make compatibility
releases for older versions, we can do so from a branch.


Keno:
Looks like femtocleaner has some things to learn here. Leaving this open as a learning opportunity.

PR #108, opened 2018-06-08
Fix all v0.7 deprecations and errors, and convert tests to testsets

rdeits:
This PR fixes all deprecation warnings and errors on v0.7-alpha and changes the minimum required Julia version to v0.7-. The only errors were caused by an ambiguity resulting from the Char -> AbstractChar change, which was easy to fix. Most of the remaining deprecation warnings were because of the removed fallback from constructors to `convert()`. In this case, I fixed them by explicitly calling `convert` where appropriate. You could, instead, define `(::Type{T<:Number})(x::Fixed) = convert(T, x)` to restore the missing constructors. I guess it's really just a matter of deciding what the "canonical" way of doing this is. 

There were also a lot of depwarns in the tests themselves because a bunch of global variables were being accidentally overwritten in for-loops. I fixed that by adding labeled testsets, which puts each test in a local scope (which should reduce the chances of one test accidentally using variables from another) and also helps document the tests themselves. I guessed at the labels, but I think I mostly got them right.


rdeits:
Alas, the github diff view really falls apart when you add a leading tab to every line in a file.... I'm going to remove the extra indentation so that the actual diff is more obvious.

REQUIRE
```diff
@@ -1,2 +1 @@
-julia 0.6
-Compat 0.35
+julia 0.7-
```
Keno: `julia 0.7-alpha`.

Keno:
Remove the deprecations while we're at it?

REQUIRE
rdeits: ok, done

rdeits:
Ok, I've removed the deprecations. The only thing we might still want to do is change some of the `convert()` methods to constructors (see https://discourse.julialang.org/t/recommended-style-for-conversion-vs-constructors-in-v0-7/11561 )

Keno:
As per that discourse thread, looks like we need to use some more constructors (and maybe define them). Other than that, this LGTM.

rdeits:
Ok, I changed all the existing `convert()` methods to constructors. As a bonus, removing the deprecations brings this package up to 100% test coverage!

PR #112, opened 2018-08-07
fix deprecations for 0.7/1.0

JeffBezanson:


src/FixedPointNumbers.jl
```diff
@@ -60,8 +60,8 @@ isinteger(x::FixedPoint{T,f}) where {T,f} = (x.i&(1<<f-1)) == 0
 # traits
 typemax(::Type{T}) where {T <: FixedPoint} = T(typemax(rawtype(T)), 0)
 typemin(::Type{T}) where {T <: FixedPoint} = T(typemin(rawtype(T)), 0)
-realmin(::Type{T}) where {T <: FixedPoint} = eps(T)
```
KristofferC: Should remove `start`, `next`, `done` on line 13?

Keno: done

PR #113, opened 2018-09-01
avoid infinite loop in promote_rule

RalphAS:
This change addresses a hang described in https://github.com/JuliaImages/ImageCore.jl/issues/64
Once the base type reaches `UInt128`, widening is an identity.

Suggestions for a good test are welcome.

### DIFF
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -159,7 +159,9 @@ end
     i1, i2 = 8*sizeof(T1)-f1, 8*sizeof(T2)-f2  # number of integer bits for each
     i = 8*sizeof(T)-f
     while i < max(i1, i2)
-        T = widen1(T)
+        Tw = widen1(T)
+        T == Tw && break
+        T = Tw
         i = 8*sizeof(T)-f
     end
     :(Normed{$T,$f})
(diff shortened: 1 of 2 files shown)


timholy:
Great catch! It could be just
```julia
promote_type(N0f8,N1f7,N2f6,N3f5,N4f4,N5f3) == Normed{UInt128,8}
```
or something.

Of course, the fact that we produce this type is a little unsatsifying. It's neither precise nor necessary:
```julia
julia> promote_type(N0f8, N1f7)
Normed{UInt16,8}     # abbreviated as N8f8

# Can all N0f8 and N1f7 numbers be represented exactly as N8f8?
julia> float(eps(N1f7))
0.007874016f0

julia> float(eps(N0f8))
0.003921569f0

julia> float(eps(N8f8))
0.003921569f0

julia> float(eps(N1f7))/float(eps(N8f8))
2.0078738f0     # are not precisely commensurate

# OK, let's check the maximum values
julia> typemax(N1f7)
2.008N1f7

julia> typemax(N0f8)
1.0N0f8

julia> typemax(N8f8)
257.0N8f8     # very excessive
```
I think it's correct to say that the smallest type that is precisely commensurate is `N8f56` (because 56 is the lcm of 7 and 8). I wonder if we should just choose `promote_type(float(T1), float(T2))`?

The existing algorithm is precise for the `Fixed` numbers, just not the `Normed` numbers.

CC @vchuravy for thoughts.

timholy:
With a test I'd be happy to merge this, and discuss what might need changing at a later point.

RalphAS:
A minimal test is in the commit hidden above.

PR #115, opened 2018-09-03
implement promotion between Fixed types

RalphAS:
Since @timholy pointed out in #113 that the promotion scheme for Normed should work for Fixed, it seems only fair to add the relevant methods.


timholy:
Thanks very much!

PR #122, opened 2019-05-15
enhance and export floattype

johnnychen94:
As a part of work in https://github.com/JuliaGraphics/ColorTypes.jl/pull/121, export `floattype` here

For Long Integers, promote them to `BigFloat` since they're not representable by `Float64` 
 ( representable in the sense of `typemax(T) <= maxintfloat(floattype(T))`)

The test is blocked by 

- [x] #121

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -23,6 +23,7 @@ export
     FixedPoint,
     Fixed,
     Normed,
+    floattype,
 # "special" typealiases
     # Q and U typealiases are exported in separate source files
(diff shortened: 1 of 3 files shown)


johnnychen94:
With the following line existed, `Base.float` looks better than `floattype`. Should we un-export `floattype` again in favor of `float`? That would be a breaking change though.

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/da3931838dd3a5f03e8ed00542a306d685002e33/src/normed.jl#L72

Issue #129, opened 2019-10-12
Rounding errors in `Normed` to `Float` conversions


kimikage:
While I was reviewing the PR #123, I found a minor bug.
```julia
julia> float(3N3f5) # this is ok
3.0f0

julia> float(3N4f12)
3.0000002f0

julia> float(3N8f8)
3.0000002f0

julia> float(3N10f6)
3.0000002f0

julia> float(3N12f4)
3.0000002f0

julia> float(3N13f3)
3.0000002f0

julia> float(3N2f6)
3.0000002f0

julia> float(3N4f4)
3.0000002f0

julia> float(3N5f3)
3.0000002f0
```
I think this is a problem with the division optimization.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/da3931838dd3a5f03e8ed00542a306d685002e33/src/normed.jl#L75-L78

The rounding errors can occur everywhere, but especially the errors at integer values are critical.
```julia
julia> isinteger(float(3N5f3))
false
```

kimikage:
I was underestimating this issue. It is not so easy to fix this without performance regression.

WIP:
```julia
function (::Type{T})(x::Normed) where {T <: AbstractFloat}
    # The following optimization for constant division may cause rounding errors.
    # y = reinterpret(x)*(one(rawtype(x))/convert(T, rawone(x)))
    # Therefore, we use a simple form here.
    y = reinterpret(x) / convert(T, rawone(x))
    convert(T, y)  # needed for types like Float16 which promote arithmetic to Float32
end
function (::Type{T})(x::Normed{Ti,f}) where {T <: Union{Float32, Float16}, Ti <: Union{UInt8, UInt16}, f}
    f == 1 && return convert(T, reinterpret(x))
    r = Int64(reinterpret(x))
    s = UInt64(1)<<(64 - 8 * sizeof(Ti))
    k = floor(Int64, s / rawone(Normed{Ti,f})) # since `rawone()>=3>2`, `r*k` is always less than 2^63.
    y = convert(Float32, r * k) * (1.0f0 / s)
    convert(T, y)
end
```
The length of the mantissa part of `Float32` is `23 (+1)` bits, so `Int64` for the intermediate variable may seem excessive. However, I think the default rounding mode (*round half to even*) may require some *bitwise hacks*, and as a result, some conditional branches may be generated in the native code.
```julia
[...]

PR #131, opened 2019-10-28
Improve accuracy of conversions from floating-point numbers (Fixes #102)

### LINKED ISSUE #102: can not convert float to the same or more number of bit fixed point
this works in Julia 0.6.2:

```julia
Float32(1.0) |> N0f8
Float32(1.0) |> N0f16
```
this will not work
```julia
Float32(1.0) |> N0f32
Float32(1.0) |> N0f64
Float64(1.0) |> N0f64
```
It looks like when the number of bits is the same or larger, it won’t work.
error message:

ArgumentError: FixedPointNumbers.Normed{UInt64,64} is a 64-bit type representing 0 values from 0.0 to 1.0; cannot represent 1.0

kimikage:
This fixes a problem with the input range checking (#102).
Since this PR includes low-level operations, special attention is required. The additional tests are for range checking, not for accuracy or integrity.
This PR has little effect on the conversions to `Normed{UInt8}` and `Normed{UInt16}`.

### DIFF
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -46,22 +46,56 @@ function Normed{T,f}(x::Normed{T2}) where {T <: Unsigned,T2 <: Unsigned,f}
 end
 N0f16(x::N0f8) = reinterpret(N0f16, convert(UInt16, 0x0101*reinterpret(x)))
 
-(::Type{U})(x::Real) where {U <: Normed} = _convert(U, rawtype(U), x)
+(::Type{U})(x::Real) where {U <: Normed} = _convert(U, x)
 
-function _convert(::Type{U}, ::Type{T}, x) where {U <: Normed,T}
-    y = round(widen1(rawone(U))*x)
-    (0 <= y) & (y <= typemax(T)) || throw_converterror(U, x)
-    U(_unsafe_trunc(T, y), 0)
+function _convert(::Type{U}, x) where {T, f, U <: Normed{T,f}}
+    if T == UInt128 # for UInt128, we can't widen
+        # the upper limit is not exact
+        (0 <= x) & (x <= (typemax(T)/rawone(U))) || throw_converterror(U, x)
+        y = round(rawone(U)*x)
+    else
+        y = round(widen1(rawone(U))*x)
+        (0 <= y) & (y <= typemax(T)) || throw_converterror(U, x)
+    end
+    reinterpret(U, _unsafe_trunc(T, y))
 end
 # Prevent overflow (https://discourse.julialang.org/t/saving-greater-than-8-bit-images/6057)
-_convert(::Type{U}, ::Type{T}, x::Float16) where {U <: Normed,T} =
-    _convert(U, T, Float32(x))
-_convert(::Type{U}, ::Type{UInt128}, x::Float16) where {U <: Normed} =
-    _convert(U, UInt128, Float32(x))
-function _convert(::Type{U}, ::Type{UInt128}, x) where {U <: Normed}
-    y = round(rawone(U)*x)   # for UInt128, we can't widen
-    (0 <= y) & (y <= typemax(UInt128)) & (x <= Float64(typemax(U))) || throw_converterror(U, x)
-    U(_unsafe_trunc(UInt128, y), 0)
+function _convert(::Type{U}, x::Float16) where {T, f, U <: Normed{T,f}}
+    if Float16(typemax(T)/rawone(U)) > Float32(typemax(T)/rawone(U))
+        x == Float16(typemax(T)/rawone(U)) && return typemax(U)
+    end
+    return _convert(U, Float32(x))
+end
+function _convert(::Type{U}, x::Tf) where {T, f, U <: Normed{T,f}, Tf <: Union{Float32, Float64}}
+    if T == UInt128 && f == 53
+        0 <= x <= Tf(3.777893186295717e22) || throw_converterror(U, x)
+    else
+        0 <= x <= Tf((typemax(T)-rawone(U))/rawone(U)+1) || throw_converterror(U, x)
+    end
+
+    significand_bits = Tf == Float64 ? 52 : 23
+    if f <= (significand_bits + 1) && sizeof(T) * 8 < significand_bits
+        return reinterpret(U, unsafe_trunc(T, round(rawone(U) * x)))
+    end
+    # cf. the implementation of `frexp`
+    Tw = f < sizeof(T) * 8 ? T : widen1(T)
+    bits = sizeof(Tw) * 8 - 1
+    xu = reinterpret(Tf == Float64 ? UInt64 : UInt32, x)
+    k = Int(xu >> significand_bits)
+    k == 0 && return zero(U) # neglect subnormal numbers
+    significand = xu | (one(xu) << significand_bits)
+    yh = unsafe_trunc(Tw, significand) << (bits - significand_bits)
+    exponent_bias = Tf == Float64 ? 1023 : 127
+    ex = exponent_bias - k + bits - f
+    yi = bits >= f ? yh - (yh >> f) : yh
+    if ex <= 0
+        ex == 0 && return reinterpret(U, unsafe_trunc(T, yi))
+        ex != -1 || signbit(signed(yi)) && return typemax(U)
+        return reinterpret(U, unsafe_trunc(T, yi + yi))
+    end
+    ex > bits && return reinterpret(U, ex == bits + 1 ? one(T) : zero(T))
+    yi += one(Tw)<<((ex - 1) & bits) # RoundNearestTiesUp
+    return reinterpret(U, unsafe_trunc(T, yi >> (ex & bits)))
 end
 
 rem(x::T, ::Type{T}) where {T <: Normed} = x
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -15,7 +15,7 @@ import Base: ==, <, <=, -, +, *, /, ~, isapprox,
 using Base: @pure
 
 # T => BaseType
-# f => Number of Bytes reserved for fractional part
+# f => Number of bits reserved for fractional part
 abstract type FixedPoint{T <: Integer, f} <: Real end
 
 
@@ -25,7 +25,7 @@ export
     Normed,
     floattype,
 # "special" typealiases
-    # Q and U typealiases are exported in separate source files
+    # Q and N typealiases are exported in separate source files
 # Functions
     scaledual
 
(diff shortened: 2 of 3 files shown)


kimikage:
On the 32-bit versions, some pre-calculated constants do not agree with the dynamically calculated values.
```julia
julia> versioninfo()
Julia Version 1.4.0-DEV.378
Commit dbaa6ff660 (2019-10-28 12:55 UTC)
Platform Info:
  OS: Windows (i686-w64-mingw32)
  WORD_SIZE: 32
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)

julia> a(x) = bitstring(x*(typemax(UInt64)/(FixedPointNumbers.rawone(N11f53))));

julia> b(x) = bitstring((x*typemax(UInt64))/(FixedPointNumbers.rawone(N11f53)));

julia> a(1)
"0100000010100000000000000000000000000000000000000000000000000000"

julia> b(1)
"0100000010100000000000000000000000000000000000000000000000000001"

julia> Float64(BigFloat(typemax(UInt64))/BigFloat(FixedPointNumbers.rawone(N11f53)))
2048.0

julia> Float64(typemax(UInt64))/Float64(FixedPointNumbers.rawone(N11f53))
2048.0000000000005

julia> BigFloat(typemax(UInt64))/BigFloat(FixedPointNumbers.rawone(N11f53))
2048.000000000000227262653140769569055940417886493182912978948283864490173983107
```
Is this a problem that should be solved by modifying the tests?

src/normed.jl
```diff
@@ -41,27 +41,63 @@ rawone(v) = reinterpret(one(v))
...
+        x == Float16(typemax(T)/rawone(U)) && return typemax(U)
+    end
+    return _convert(U, Float32(x))
+end
+function _convert(::Type{U}, x::Tf) where {T, f, U <: Normed{T,f}, Tf <: Union{Float32, Float64}}
+    if T == UInt128 && f == 53
+        0 <= x <= Tf(3.777893186295717e22) || throw_converterror(U, x)
+    else
+        0 <= x <= Tf((typemax(T)-rawone(U))/rawone(U)+1) || throw_converterror(U, x)
+    end
```
kimikage: `(typemax(T)-rawone(U))/rawone(U)+1` may have higher accuracy than `typemax(T)/rawone(U)` because the `1`s in the dividend are reduced. However, the divisor in `Float64` still have rounding errors when `f >= 53`. So, the specialization for `Normed{UInt128, 53}` is needed. FYI, using `BigFloat` hinders the constant pre-calculation and inline expansion.

Although this can be a temporary measure, I think the problem with pre-calculating constants is potentially troublesome.

kimikage:
To avoid impact on the performance, this PR does not change the `rem`. 
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/da3931838dd3a5f03e8ed00542a306d685002e33/src/normed.jl#L67-L70

Therefore, this does not fix the problems such as:
```julia
julia> using Colors

julia> white = parse(RGB{Float32}, "white")
RGB{Float32}(1.0f0,1.0f0,1.0f0)

julia> convert(RGB{N0f32}, white)
RGB{N0f32}(0.0,0.0,0.0)
```

src/normed.jl
```diff
@@ -41,27 +41,63 @@ rawone(v) = reinterpret(one(v))
...
+    k == 0 && return zero(U) # neglect subnormal numbers
+    significand = xu | (one(xu) << significand_bits)
+    yh = unsafe_trunc(Tw, significand) << (bits - significand_bits)
+    exponent_bias = Tf == Float64 ? 1023 : 127
+    ex = exponent_bias - k + bits - f
+    if ex <= 0
+        yi = yh - (yh >> f)
+        ex == 0 && return reinterpret(U, unsafe_trunc(T, yi))
+        if ex == -1
+            (yi & (one(Tw) << bits)) != 0 && return typemax(U)
```
kimikage: This is shorter in the source code and probably shorter in the native code too. However, the speed will not change.
```suggestion
            signbit(signed(yi)) && return typemax(U)
```

src/normed.jl
```diff
@@ -41,27 +41,61 @@ rawone(v) = reinterpret(one(v))
...
+        x == Float16(typemax(T)/rawone(U)) && return typemax(U)
+    end
+    return _convert(U, Float32(x))
+end
+function _convert(::Type{U}, x::Tf) where {T, f, U <: Normed{T,f}, Tf <: Union{Float32, Float64}}
+    if T == UInt128 && f == 53
+        0 <= x <= Tf(3.777893186295717e22) || throw_converterror(U, x)
+    else
+        0 <= x <= Tf((typemax(T)-rawone(U))/rawone(U)+1) || throw_converterror(U, x)
+    end
```
kimikage: `(typemax(T)-rawone(U))/rawone(U)+1` may have slightly higher accuracy than `typemax(T)/rawone(U)` because the `1`s in the dividend are reduced.  However, this have no effect for `UInt128`. So, the specialization for `Normed{UInt128, 53}` is needed (`f == 53` is one of singular points). FYI, using `BigFloat` hinders the constant pre-calculation and inline expansion.

Although this can be a temporary measure, I think the problem with pre-calculating constants is potentially troublesome.

src/normed.jl
```diff
@@ -41,27 +41,61 @@ rawone(v) = reinterpret(one(v))
 function Normed{T,f}(x::Normed{T2}) where {T <: Unsigned,T2 <: Unsigned,f}
     U = Normed{T,f}
     y = round((rawone(U)/rawone(x))*reinterpret(x))
-    (0 <= y) & (y <= typemax(T)) || throw_converterror(U, x)
+    0 <= y <= typemax(T) || throw_converterror(U, x)
```
timholy: These are actually subtly different, and at least at one point there was a performance difference (doesn't seem to be one now, at least not on my current machine):
```julia
julia> function f1(x)
           0.25 <= x <= 0.75 || throw(DomainError(x))
           return x
       end
f1 (generic function with 1 method)

julia> function f2(x)
           ((0.25 <= x) & (x <= 0.75)) || throw(DomainError(x))
           return x
       end
f2 (generic function with 1 method)

julia> @code_lowered f1(0.5)
CodeInfo(
1 ─ %1 = 0.25 <= x
└──      goto #3 if not %1
2 ─      @_3 = x <= 0.75
└──      goto #4
3 ─      @_3 = false
4 ┄ %6 = @_3
└──      goto #6 if not %6
5 ─      goto #7
6 ─ %9 = Main.DomainError(x)
└──      Main.throw(%9)
7 ┄      return x
)

julia> @code_lowered f2(0.5)
CodeInfo(
1 ─ %1 = 0.25 <= x
│   %2 = x <= 0.75
│   %3 = %1 & %2
└──      goto #3 if not %3
2 ─      goto #4
3 ─ %6 = Main.DomainError(x)
└──      Main.throw(%6)
4 ┄      return x
)

julia> using BenchmarkTools

julia> @btime f1(x) setup=(x=rand()/10 + 0.3)
  1.980 ns (0 allocations: 0 bytes)
0.34079130696459065

julia> @btime f2(x) setup=(x=rand()/10 + 0.3)
  1.977 ns (0 allocations: 0 bytes)
0.30934306198706746
```

kimikage: Yes. I guess, using `&` instead of `&&` may reduce condition branches. However, `throw_converterror(U, x)` is annotated with `@noinline` and throws the exception. This means that we cannot avoid condition branches, and this is not SIMD-suitable. Depending on the type (and CPU), the comparison with zero can be specialized (this is the reason why I did not use `zero()`).

Of course, I follow the law of *"if it works, don't touch it"*.:smile:

kimikage:
> Does this have negative consequences for performance?

Unfortunately, my *software* floating-point calculation is ~2.5x slower.
The following are the benchmark results based on the (inverse) `Vec4` version of https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-544244496

| x86_64              |    before |     after |
|:--------------------|:---------:|:---------:|
| `Float32`->`N0f8`   | 15.999 μs | 14.099 μs |
| `Float32`->`N8f24`  | 20.000 μs | 48.199 μs |
| `Float32`->`N0f32`  | 20.100 μs | 49.100 μs |
| `Float64`->`N0f8`   | 15.999 μs | 13.999 μs |
| `Float64`->`N0f32`  | 16.099 μs | 14.299 μs |
| `Float64`->`N11f53` | 16.099 μs | 14.299 μs |
| `Float64`->`N0f64`  | 40.999 μs | 68.900 μs |

However, we cannot measure the conversion speed of what cannot be converted. As implied above, in applications where the speed is needed, `rem` is used and the conversions with range checking should not be used.

src/normed.jl
timholy: Yeah, you definitely need one branch, but one is usually better than two. For example when Matt & I were developing the array infrastructure we noticed it was better to check bounds on all indexes and combine with `&`, and then throw a `BoundsError` if necessary, rather than having one branch per dimension.

But this is definitely something that depends on the CPU and its ability to predict branches, so it can even be hard to benchmark. I would generally suspect that fewer branches is a good thing.

timholy:
Thanks for posting those numbers. Correctness is infinitely more important than speed, so I would support merging this.

src/normed.jl
kimikage: That's right. Nevertheless, I think it is another good practice to avoid something tricky and leave it to the compiler (LLVM) *now*, because this is *usually* not a pair of integer comparisons.

However, I do never intend to dispute about the comparison manner, which has no definitive answer. Instead I would care about the `/`, but it is another matter.

kimikage:
For the lines based on the current source code, I reverted the comparison manner. (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/pull/131#discussion_r340806823)

In the new conversion method, I used the `a <= x <= b` style. This decision is based on [my policy](https://github.com/JuliaMath/FixedPointNumbers.jl/pull/131#discussion_r341069209) and the rough benchmark. Honestly, the reason is that it is shorter in source codes. :smiley:
Although the inconsistency makes me somewhat itchy, I think it will also make more people notice this comparison issue.

timholy:
I'm fine with whatever you choose here. Very grateful for this change!

kimikage:
I am ready to rebase and merge this PR, but not yet ready for other issues. When When do you bump the version number up?

timholy:
Version numbers are cheap: any time it's potentially breaking, we should bump the minor version number. If the other fixes will follow quickly I'd say let's consolidate the churn and wait to tag a release until it's all done. If not, let's get a release out for the fixes we have.

As the community has moved to putting [upper version bounds on packages](https://discourse.julialang.org/t/announcement-automatic-merging-for-the-general-registry/29961), it's also more likely that breaking changes won't break things for users.

kimikage:
If this fix causes problems, consider using `rem` (`% Nxfy`) instead.

PR #135, opened 2019-11-07
Remove `testapprox` dependency and Reduce the number of test cases for display (#133)

### LINKED ISSUE #133: `testapprox` should be moved or inlined
The "fixed" testset depends on `testapprox` which is defined in "test/normed.jl" (formerly named ufixed.jl).
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/70ae1d6ed27c7005191f23b56343e98ff9ef2026/test/fixed.jl#L86-L90

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/70ae1d6ed27c7005191f23b56343e98ff9ef2026/test/normed.jl#L233-L240

This thwarts the selective execution of tests and causes an error(`UndefVarError: testapprox not defined`) in certain environments (e.g. Julia v1.0.3 32-bit ARMv7).

I think `testapprox` is not so complicated. It may be a good idea to inline the function or define it locally.
This may be off topic, but I also think there are too many test cases generated by `testapprox` (or its friend `testtrunc`). `@test` should be outside the loop.

kimikage:
This fixes #133.
This does not change the actual test cases except one *wraparound* case, and also does not change `testtrunc`.

### DIFF
--- a/test/fixed.jl
+++ b/test/fixed.jl
@@ -84,8 +84,11 @@ end
 end
 
 @testset "testapprox" begin
-    for T in [Fixed{Int8,7}, Fixed{Int16,8}, Fixed{Int16,10}]
-        testapprox(T)  # defined in ufixed.jl
+    @testset "approx $T" for T in [Fixed{Int8,7}, Fixed{Int16,8}, Fixed{Int16,10}]
+        xs = typemin(T):eps(T):typemax(T)-eps(T)
+        @test all(x -> x ≈ x + eps(T), xs)
+        @test all(x -> x + eps(T) ≈ x, xs)
+        @test !any(x -> x - eps(T) ≈ x + eps(T), xs)
     end
 end
 
--- a/test/normed.jl
+++ b/test/normed.jl
@@ -230,18 +230,12 @@ end
     end
 end
 
-function testapprox(::Type{T}) where {T}
-    for x = typemin(T):eps(T):typemax(T)-eps(T)
-        y = x+eps(T)
-        @test x ≈ y
-        @test y ≈ x
-        @test !(x ≈ y+eps(T))
-    end
-end
-
 @testset "approx" begin
-    for T in FixedPointNumbers.UF
-        testapprox(T)
+    @testset "approx $T" for T in FixedPointNumbers.UF
+        xs = typemin(T):eps(T):typemax(T)-eps(T)
+        @test all(x -> x ≈ x + eps(T), xs)
+        @test all(x -> x + eps(T) ≈ x, xs)
+        @test !any(x -> x - eps(T) ≈ x + eps(T), xs)
     end
 end
 


johnnychen94:
> This thwarts the selective execution of tests and causes an error(UndefVarError: testapprox not defined) in certain environments (e.g. Julia v1.0.3 32-bit ARMv7).

I guess this is caused by the outmost test group `normed` and `fixed`

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/70ae1d6ed27c7005191f23b56343e98ff9ef2026/test/runtests.jl#L5-L10

So moving these functions outside of it might fix this issue. For example, moving them into `test/utils.jl`

kimikage:
> Just breaking this function and doing copy&paste things doesn't look good to me...

I generally agree with you. However, the function is not a so-called utility function and is nothing but an implementation of tests.
If some tests can be shared between `Fixed` and `Normed`, it means that they are tests for `FixedPoint`, which specify the behavior of `FixedPoint`. I think what we should share is not the content, but the outside (i.e. `testset`). However, it goes against the current style of providing independent testsets of `Fixed` and` Normed`.

This is off topic, but I think that sharing the implementation should be prioritized over sharing the test.
https://github.com/JuliaMath/FixedPointNumbers.jl/compare/master...kimikage:commonize

johnnychen94:
It makes sense.

I don't have a strong opinion on this, and just want to point out an alternative here. I believe you're aware of what you're doing. 😄 Thank you for digging into the details and taking care of this package.

kimikage:
I am a little educated in electrics, but not educated in computer science or image processing.:sweat_smile:
> If I understand it correctly, `testapprox` is a local function in group `normed`, using it in `fixed` *should* throw an undef error.

As a matter of fact, I have not clarified *the cause*. I got tired of seeing "Error During Test at test/normed.jl" while debugging, but have never seen the "UndefVarError" on my x86 machines.
In fact, `testapprox` has no `local` keyword. 
> Notably missing from this table are `begin` blocks and `if` blocks which do *not* introduce new scopes.

https://docs.julialang.org/en/v1/manual/variables-and-scoping/

Of course, whatever the cause, your alternative plan should solve the problem correctly!

johnnychen94:
Hmmm, actually `begin...end` block does introduce a new scope, at least the following codes says so:

```julia
julia> using Test

julia> @testset "Foo" begin
           function foo(x)
               println(x)
               return x
           end
       end
Test Summary: |
Foo           | No tests
Test.DefaultTestSet("Foo", Any[], 0, false)

julia> foo
ERROR: UndefVarError: foo not defined
```

johnnychen94:
> Notably missing from this table are begin blocks and if blocks which do not introduce new scopes.

Hmmm, yes it is, but here it's `@testset` that introduces a new scope here,

```julia
# test.jl
using Test

@testset "Foo" begin
    function foo(x)
        println(x)
        return x
    end
    foo(1)
end

foo(2)

```

```bash
» julia --startup=no test.jl 
1
Test Summary: |
Foo           | No tests
ERROR: LoadError: UndefVarError: foo not defined
Stacktrace:
 [1] top-level scope at /Users/jc/Desktop/test.jl:11
 [2] include at ./boot.jl:328 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1094
 [4] include(::Module, ::String) at ./Base.jl:31
 [5] exec_options(::Base.JLOptions) at ./client.jl:295
 [6] _start() at ./client.jl:464
in expression starting at /Users/jc/Desktop/test.jl:11
```

Actually, I'm a little surprised that the current codes pass the test...

kimikage:
I am sorry for misleading you.
`@testset` encloses the tests in a `try` block, i.e. introduces a new scope. Usually, we don't have to be aware of this, but `include` works with the (implicitly) specified `Module`.
That is all I understand about this. :disappointed_relieved:

kimikage:
I submitted a new issue #139, but this PR is not staled. You may merge or close this PR if you want to.

timholy:
Thanks!

PR #136, opened 2019-11-11
Fix `rem` problem with negative float inputs on ARMv7 (Fixes #134)

### LINKED ISSUE #134: `rem` returns zero for negative float input on ARMv7
To check the modification for the issue #129, I ran the tests on a 32-bit ARMv7 system (RPi 2 Model B v1.2). And then, I faced a problem with `rem` (`%`).
```julia
modulus: Test Failed at ~/.julia/dev/FixedPointNumbers/test/normed.jl:148
  Expression: (-0.3 % N0f8).i == round(Int, -0.3 * 255) % UInt8
   Evaluated: 0x00 == 0xb4
modulus: Test Failed at ~/.julia/dev/FixedPointNumbers/test/normed.jl:154
  Expression: (-0.3 % N6f10).i == round(Int, -0.3 * 1023) % UInt16
   Evaluated: 0x0000 == 0xfecd
```
The cause is the behavior of `unsafe_trunc`.

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/70ae1d6ed27c7005191f23b56343e98ff9ef2026/src/normed.jl#L103

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/70ae1d6ed27c7005191f23b56343e98ff9ef2026/src/normed.jl#L204-L205

```julia
julia> versioninfo()
Julia Version 1.0.3
Platform Info:
  WORD_SIZE: 32
  LIBM: libopenlibm

julia> unsafe_trunc(UInt8, -76.0) # or the intrinsic `fptoui`
0x00

julia> unsafe_trunc(Int8, -76.0)
-76

julia> unsafe_trunc(UInt8, unsafe_trunc(Int8, -76.0))
0xb4
```
(The problem occurs not only on v1.0.3 but also on v1.0.5 and v1.2.0. I have not tried the 64-bit.)

Although the behavior of `unsafe_trunc` may not be what we want, this is not a bug.
> If the value is not representable by `T`, an arbitrary value will be returned.

https://docs.julialang.org/en/v1/base/math/#Base.unsafe_trunc

However, I don't think it is good to make the `rem` users aware of the *internal* `unsafe_trunc`.
The workaround is to convert the value to `Signed` temporarily as shown above.

BTW, the behavior of `Normed`'s `rem`, which is specified by the above tests seems to be not intuitive. (Since I know the inside of `Normed`, I think the behavior is reasonable, though.)
So, I think it is another option to eliminate the tests for negative float inputs, i.e. make it an undefined behavior.

kimikage:
This fixes issue #134.

### DIFF
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -203,3 +203,9 @@ end
 
 _unsafe_trunc(::Type{T}, x::Integer) where {T} = x % T
 _unsafe_trunc(::Type{T}, x) where {T}          = unsafe_trunc(T, x)
+if !signbit(signed(unsafe_trunc(UInt, -12.345)))
+    # a workaround for 32-bit ARMv7 (issue #134)
+    function _unsafe_trunc(::Type{T}, x::AbstractFloat) where {T}
+        unsafe_trunc(T, unsafe_trunc(typeof(signed(zero(T))), x))
+    end
+end


kimikage:
My initial plan was to fix #133 first, but I prefer to pass the "Normed" tests first. I think it does not fix #133 radically, but may avoid #133.

Issue #130, opened 2019-10-23
What is `scaledual`?


kimikage:
I am working on improving the accuracy of the conversions from `Normed` to `Float` (#129), and I am interested in `scaledual`, which seems to be related with the conversions.

The `scaledual` was introduced in https://github.com/JuliaMath/FixedPointNumbers.jl/commit/d1087f71e20be5e1ee66511b6a3ed9726b7646ac.

> The introduction of `scaledual` is a bit speculative, but it can essentially double the speed of certain operations. It has the following property:
>
>``` julia
>bd, ad = scaledual(b, a)
>b*a == bd*ad
>```
>
>but the RHS might be faster (particularly for floating-point `b` and an array `a` of fixed-point numbers).
>

_Originally posted by @timholy in https://github.com/JuliaMath/FixedPointNumbers.jl/pull/2#issuecomment-50288877_

However, in the current codebase, I think `scaledual` does not have the property above as its test specifies. (`a[1] != af8[1]`)
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/da3931838dd3a5f03e8ed00542a306d685002e33/test/normed.jl#L287-L294

I do my best for #129, but a slowdown is inevitable. If `scaledual` is helpful as a workaround for people who prefer speed over accuracy, I feel relieved.

@timholy , did I not understand that correctly?

timholy:
Dang, you're right. Obviously `scaledual` has not gotten much/any use. (I haven't used it myself even though I should sometimes.)

I will submit a fix, but it will be a breaking change so we'll need to do a minor-version-number bump when we release it.

kimikage:
Apart from the appearance:sweat_smile:, the new `scaledual`(#132) finely works as same as the current floating-point conversion! :tada:

[Benchmark](https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-544244496)
```julia
function Vec3{T}(v::Vec3{U}, _) where {T, U}
    x::T = *(scaledual(T, v.x)...)
    y::T = *(scaledual(T, v.y)...)
    z::T = *(scaledual(T, v.z)...)
    Vec3{T}(x, y, z)
end
function Vec4{T}(v::Vec4{U}, _) where {T, U}
    x::T = *(scaledual(T, v.x)...)
    y::T = *(scaledual(T, v.y)...)
    z::T = *(scaledual(T, v.z)...)
    w::T = *(scaledual(T, v.w)...)
[...]

Issue #125, opened 2019-06-04
poor performance


johnnychen94:
Actually, this surprises me a bit, any hints/ideas on how to fix this?

A similar result can be observed for `*`,`.*`, and etc.

```julia
julia> x_n0f8 = rand(N0f8, 1000, 1000);

julia> x_float64 = rand(Float64, 1000, 1000);

julia> x_uint8 = rand(UInt8, 1000, 1000);

julia> @benchmark x_n0f8 .+ x_n0f8
BenchmarkTools.Trial: 
  memory estimate:  976.84 KiB
  allocs estimate:  6
  --------------
  minimum time:     3.547 ms (0.00% GC)
  median time:      3.732 ms (0.00% GC)
  mean time:        3.838 ms (1.11% GC)
  maximum time:     7.032 ms (0.00% GC)
  --------------
  samples:          1301
  evals/sample:     1

julia> @benchmark x_float64 .+ x_float64
BenchmarkTools.Trial: 
  memory estimate:  7.63 MiB
  allocs estimate:  4
  --------------
  minimum time:     1.214 ms (0.00% GC)
  median time:      1.346 ms (0.00% GC)
  mean time:        1.869 ms (29.19% GC)
  maximum time:     5.726 ms (71.81% GC)
  --------------
  samples:          2667
  evals/sample:     1

julia> @benchmark x_uint8 .+ x_uint8
BenchmarkTools.Trial: 
  memory estimate:  976.75 KiB
  allocs estimate:  4
  --------------
  minimum time:     80.220 μs (0.00% GC)
  median time:      87.781 μs (0.00% GC)
  mean time:        235.548 μs (18.39% GC)
  maximum time:     1.844 ms (69.85% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

timholy:
I am surprised too. Best approach would be to profile it and see where it's spending time. Or step into it with a debugger and see which version of `+` is getting called.

kimikage:
One reason why `x_float64 .+ x_float64` is faster, is that the SIMD optimization is relatively well carried out. However, I don't know yet why `x_n0f8 .+ x_n0f8` is slow.

Although it is not directly related to the subject of this issue, I think `*` for `Normed` is inefficient since it converts the parameters into Float and converts the multiplied back to UInt.
[...]

PR #138, opened 2019-11-17
Improve accuracy of conversions to floating-point numbers (#129)

kimikage:
This fixes (mitigates) the issue #129.

### DIFF
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -105,11 +105,135 @@ rem(x::Float16, ::Type{T}) where {T <: Normed} = rem(Float32(x), T)  # avoid ove
 
 float(x::Normed) = convert(floattype(x), x)
 
-Base.BigFloat(x::Normed) = reinterpret(x)*(1/BigFloat(rawone(x)))
+macro f32(x::Float64) # just for hexadecimal floating-point literals
+    :(Float32($x))
+end
+macro exp2(n)
+     :(_exp2(Val($(esc(n)))))
+end
+_exp2(::Val{N}) where {N} = exp2(N)
+
+# for Julia v1.0, which does not fold `div_float` before inlining
+inv_rawone(x) = (@generated) ? (y = 1.0 / rawone(x); :($y)) : 1.0 / rawone(x)
+
 function (::Type{T})(x::Normed) where {T <: AbstractFloat}
-    y = reinterpret(x)*(one(rawtype(x))/convert(T, rawone(x)))
+    # The following optimization for constant division may cause rounding errors.
+    # y = reinterpret(x)*(one(rawtype(x))/convert(T, rawone(x)))
+    # Therefore, we use a simple form here.
+    # If you prefer speed over accuracy, consider using `scaledual` instead.
+    y = reinterpret(x) / convert(promote_type(T, floattype(x)), rawone(x))
     convert(T, y)  # needed for types like Float16 which promote arithmetic to Float32
 end
+
+function Base.Float16(x::Normed{Ti,f}) where {Ti <: Union{UInt8, UInt16, UInt32}, f}
+    f == 1 ? Float16(x.i) : Float16(Float32(x))
+end
+function Base.Float16(x::Normed{Ti,f}) where {Ti <: Union{UInt64, UInt128}, f}
+    f == 1 ? Float16(x.i) : Float16(Float64(x))
+end
+
+function Base.Float32(x::Normed{UInt8,f}) where f
+    f == 1 && return Float32(x.i)
+    f == 2 && return Float32(Int32(x.i) * 0x101) * @f32(0x550055p-32)
+    f == 3 && return Float32(Int32(x.i) * 0x00b) * @f32(0xd4c77bp-30)
+    f == 4 && return Float32(Int32(x.i) * 0x101) * @f32(0x110011p-32)
+    f == 5 && return Float32(Int32(x.i) * 0x003) * @f32(0xb02c0bp-30)
+    f == 6 && return Float32(Int32(x.i) * 0x049) * @f32(0xe40039p-36)
+    f == 7 && return Float32(Int32(x.i) * 0x01f) * @f32(0x852b5fp-35)
+    f == 8 && return Float32(Int32(x.i) * 0x155) * @f32(0xc0f0fdp-40)
+    0.0f0
+end
+function Base.Float32(x::Normed{UInt16,f}) where f
+    f32 = Float32(x.i)
+    f ==  1 && return f32
+    f ==  2 && return f32 * @f32(0x55p-8)  + f32 * @f32(0x555555p-32)
+    f ==  3 && return f32 * @f32(0x49p-9)  + f32 * @f32(0x249249p-33)
+    f ==  4 && return f32 * @f32(0x11p-8)  + f32 * @f32(0x111111p-32)
+    f ==  5 && return f32 * @f32(0x21p-10) + f32 * @f32(0x108421p-35)
+    f ==  6 && return f32 * @f32(0x41p-12) + f32 * @f32(0x041041p-36)
+    f ==  7 && return f32 * @f32(0x81p-14) + f32 * @f32(0x204081p-42)
+    f == 16 && return f32 * @f32(0x01p-16) + f32 * @f32(0x010001p-48)
+    Float32(x.i / rawone(x))
+end
+function Base.Float32(x::Normed{UInt32,f}) where f
+    f == 1 && return Float32(x.i)
+    i32 = unsafe_trunc(Int32, x.i)
+    if f == 32
+        rh, rl = Float32(i32>>>16), Float32((i32&0xFFFF)<<8 | (i32>>>24))
+        return muladd(rh, @f32(0x1p-16), rl * @f32(0x1p-40))
+    elseif f >= 25
+        rh, rl = Float32(i32>>>16),Float32(((i32&0xFFFF)<<14) + (i32>>>(f-14)))
+        return muladd(rh, Float32(@exp2(16-f)), rl * Float32(@exp2(-14-f)))
+    end
+    # FIXME: avoid the branch in native x86_64 (non-SIMD) codes
+    m = ifelse(i32 < 0, 0x1p32 * inv_rawone(x), 0.0)
+    Float32(muladd(Float64(i32), inv_rawone(x), m))
+end
+function Base.Float32(x::Normed{Ti,f}) where {Ti <: Union{UInt64, UInt128}, f}
+    f == 1 ? Float32(x.i) : Float32(Float64(x))
+end
+
+function Base.Float64(x::Normed{Ti,f}) where {Ti <: Union{UInt8, UInt16}, f}
+    Float64(Normed{UInt32,f}(x))
+end
+function Base.Float64(x::Normed{UInt32,f}) where f
+    f64 = Float64(x.i)
+    f ==  1 && return f64
+    f ==  2 && return (f64 * 0x040001) * 0x15555000015555p-72
+    f ==  3 && return (f64 * 0x108421) * 0x11b6db76924929p-75
+    f ==  4 && return (f64 * 0x010101) * 0x11000011000011p-72
+    f ==  5 && return (f64 * 0x108421) * 0x04000002000001p-75
+    f ==  6 && return (f64 * 0x09dfb1) * 0x1a56b8e38e6d91p-78
+    f ==  7 && return (f64 * 0x000899) * 0x0f01480001e029p-70
+    f ==  8 && return (f64 * 0x0a5a5b) * 0x18d300000018d3p-80
+    f ==  9 && return (f64 * 0x001001) * 0x080381c8e3f201p-72
+    f == 10 && return (f64 * 0x100001) * 0x04010000000401p-80
+    f == 11 && return (f64 * 0x000009) * 0x0e3aaae3955639p-66
+    f == 12 && return (f64 * 0x0a8055) * 0x186246e46e4cfdp-84
+    f == 13 && return (f64 * 0x002001) * 0x10000004000001p-78
+    f == 14 && return (f64 * 0x03400d) * 0x13b13b14ec4ec5p-84
+    f == 15 && return (f64 * 0x000259) * 0x06d0c5a4f3a5e9p-75
+    f == 16 && return (f64 * 0x011111) * 0x00f000ff00fff1p-80
+    f == 18 && return (f64 * 0x0b06d1) * 0x17377445dd1231p-90
+    f == 19 && return (f64 * 0x080001) * 0x00004000000001p-76
+    f == 20 && return (f64 * 0x000101) * 0x0ff010ef10ff01p-80
+    f == 21 && return (f64 * 0x004001) * 0x01fff8101fc001p-84
+    f == 22 && return (f64 * 0x002945) * 0x18d0000000018dp-88
+    f == 23 && return (f64 * 0x044819) * 0x07794a23729429p-92
+    f == 27 && return (f64 * 0x000a21) * 0x0006518c7df9e1p-81
+    f == 28 && return (f64 * 0x00000d) * 0x13b13b14ec4ec5p-84
+    f == 30 && return (f64 * 0x001041) * 0x00fc003f03ffc1p-90
+    f == 32 && return (f64 * 0x010101) * 0x00ff0000ffff01p-96
+    f64 / rawone(x)
+end
+function Base.Float64(x::Normed{UInt64,f}) where f
+    f == 1 && return Float64(x.i)
+    if f >= 53
+        rh = Float64(unsafe_trunc(Int64, x.i >> 16)) * @exp2(16-f) # upper 48 bits
+        rl = Float64(unsafe_trunc(Int32, x.i&0xFFFF)) * @exp2(-f)  # lower 16 bits
+        return rh + muladd(rh, @exp2(-f), rl)
+    end
+    x.i / rawone(x)
+end
+function Base.Float64(x::Normed{UInt128,f}) where f
+    f == 1 && return Float64(x.i)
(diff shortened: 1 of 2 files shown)


kimikage:
## Benchmark
see:
https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-544244496
https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-553185335
https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-554698883

## Known Issues for Users
### Rounding error
In some cases, the 1-lsb errors still occurs. (see also: https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-544119236)

For example:
```julia
julia> x = reinterpret(N4f28, 0b1111_1111_1111_1111_1111_1111_0111_0000)
15.999999523N4f28

julia> Float32(x)
16.0f0

julia> Float32(BigFloat(Rational(x)))
15.999999f0
```

**Workaround**: use of the divisions.

### Slowdown
With a few exceptions, the accuracy comes at price of the slowdown.
The following particular cases are relatively noticeable.
- `N8f8`, `N7f9`, `N6f10`, `N5f11`, `N4f12`, `N3f13`, `N2f14`, `N1f15` -> `Float32`
N16

- `N15f17`, `N8f24`, `N7f25`, `N6f26`, `N3f29`, `N1f31` -> `Float64`
N32

**Workaround**: use of `scaledual`. (see also: #130)


### Slowdown with `ReinterpretArray`
The conversion methods are designed to be suitable for the SIMD vectorization. However, the current `ReinterpretArray` is not SIMD-suitable (cf. #125, https://github.com/JuliaLang/julia/issues/28980), so it will cause a further slowdown.
Especially, in cases of `Normed{UInt32,f}`->`Float32` where `2 <= f <= 24`, the LLVM generates a destructive conditional branch.

**Workaround**: use of `Array` instead of `ReinterpretArray` (i.e. avoiding `rand(::Type{<:Normed}, sz::Dims)`)

### Conversions to `BigFloat`
This PR does not change the conversions to `BigFloat`. I think the `BigFloat` users need the accuracy rather than the speed.

**Workaround**: use of `Rational` in the first step of two-step conversion.

src/normed.jl
```diff
@@ -105,11 +105,135 @@ rem(x::Float16, ::Type{T}) where {T <: Normed} = rem(Float32(x), T)  # avoid ove
...
+    # FIXME: avoid the branch in native x86_64 (non-SIMD) codes
+    m = ifelse(i32 < 0, 0x1p32 * inv_rawone(x), 0.0)
+    Float32(muladd(Float64(i32), inv_rawone(x), m))
+end
+function Base.Float32(x::Normed{Ti,f}) where {Ti <: Union{UInt64, UInt128}, f}
+    f == 1 ? Float32(x.i) : Float32(Float64(x))
+end
+
+function Base.Float64(x::Normed{Ti,f}) where {Ti <: Union{UInt8, UInt16}, f}
+    Float64(reinterpret(Normed{UInt32,f}, UInt32(x.i)))
```
kimikage: I think the simple (overflow-free) widening is worth specializing. It may be helpful for the accumulation (`sum`) without floating-point conversions.

src/normed.jl
```diff
@@ -105,11 +105,135 @@ rem(x::Float16, ::Type{T}) where {T <: Normed} = rem(Float32(x), T)  # avoid ove
...
 float(x::Normed) = convert(floattype(x), x)
 
-Base.BigFloat(x::Normed) = reinterpret(x)*(1/BigFloat(rawone(x)))
+macro f32(x::Float64) # just for hexadecimal floating-point literals
+    :(Float32($x))
+end
+macro exp2(n)
+     :(_exp2(Val($(esc(n)))))
+end
+@generated _exp2(::Val{N}) where {N} = (x = exp2(N); :($x))
```
timholy: Could both of these be `Base.@pure`? With `@generated` one should technically provide a fallback (e.g., https://github.com/JuliaLang/julia/blob/e5ee8da70a646f1f54730e727932ebaeaee256d5/base/namedtuple.jl#L75-L83) in preparation for the day when it becomes more common to ship precompiled binaries.

kimikage: Oh! `_exp2(::Val{N}) where {N} = exp2(N)` seems to work well! (not fully tested)
I found that the generated function can be a workaround for the integer division problems. And then, I thought the same technique could be used for `exp2`, and I did it as shown above.

However, I don't know a substitution for the *generated* `inv_rawone()`.

kimikage: Although I do not fully understand the mechanism of `@generated`, I added a fallback by following the example.
```julia
inv_rawone(x) = (@generated) ? (y = 1.0 / rawone(x); :($y)) : 1.0 / rawone(x)
```

BTW, I am thinking of moving these low-level utility functions and macros to a new file (e.g. "src/utilities.jl")
Surprisingly, `rawone` is defined in "src/normed.jl". Moreover, although there is no noticeable overhead, the implementation of `rawone` takes a *detour*:
```julia
interpret> FixedPointNumbers.rawone(1N5f3)
rawone(v) in FixedPointNumbers at ~/.julia/packages/FixedPointNumbers/870tH/src/normed.jl:38
 one(x::Normed) in FixedPointNumbers at ~/.julia/packages/FixedPointNumbers/870tH/src/normed.jl:37
  oneunit(x::Normed) in FixedPointNumbers at ~/.julia/packages/FixedPointNumbers/870tH/src/normed.jl:36
   one(::Type{T}) where T<:Normed in FixedPointNumbers at ~/.julia/packages/FixedPointNumbers/870tH/src/normed.jl:34
    oneunit(::Type{T}) where T<:Normed in FixedPointNumbers at ~/.julia/packages/FixedPointNumbers/870tH/src/normed.jl:32
  #unused# = Normed{UInt8,3}
  T = Normed{UInt8,3}
 31  function oneunit(::Type{T}) where T <: Normed
 32      T(typemax(rawtype(T)) >> (8 * sizeof(T) - nbitsfrac(T)), 0)
     end
```
Even though we can get the *`rawone`* value (which `isa Unsigned`) in `oneunit(::Type{T})`, it will be converted to `Normed`, and reinterpreted back to `Unsigned`.

Should I refactor the `zero`, `one`, `oneunit` and `rawone` first, as a part of #139? Or keep the utility functions in "src/normed.jl" for the moment?

src/normed.jl
```diff
@@ -105,11 +105,135 @@ rem(x::Float16, ::Type{T}) where {T <: Normed} = rem(Float32(x), T)  # avoid ove
...
+macro f32(x::Float64) # just for hexadecimal floating-point literals
+    :(Float32($x))
+end
+macro exp2(n)
+     :(_exp2(Val($(esc(n)))))
+end
+_exp2(::Val{N}) where {N} = exp2(N)
+
+# for Julia v1.0, which does not fold `div_float` before inlining
+inv_rawone(x) = (@generated) ? (y = 1.0 / rawone(x); :($y)) : 1.0 / rawone(x)
```
timholy: With
```diff
diff --git a/src/normed.jl b/src/normed.jl
index 737657b..07f5b86 100644
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -114,7 +114,7 @@ end
 _exp2(::Val{N}) where {N} = exp2(N)
 
 # for Julia v1.0, which does not fold `div_float` before inlining
-inv_rawone(x) = (@generated) ? (y = 1.0 / rawone(x); :($y)) : 1.0 / rawone(x)
+@pure inv_rawone(x) = 1.0 / rawone(x)
 
 function (::Type{T})(x::Normed) where {T <: AbstractFloat}
     # The following optimization for constant division may cause rounding errors.
```
I get
```julia
julia> x = N8f24(15.2)
15.2N8f24

julia> @code_llvm Float32(x)

;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:159 within `Float32'
define float @julia_Float32_17656(%jl_value_t addrspace(10)*, { i32 } addrspace(11)* nocapture nonnull readonly dereferenceable(4)) {
top:
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:160 within `Float32'
; ┌ @ Base.jl:20 within `getproperty'
   %2 = getelementptr inbounds { i32 }, { i32 } addrspace(11)* %1, i64 0, i32 0
; └
; ┌ @ int.jl:498 within `unsafe_trunc'
; │┌ @ int.jl:464 within `rem'
    %3 = load i32, i32 addrspace(11)* %2, align 4
; └└
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:169 within `Float32'
; ┌ @ promotion.jl:349 within `<' @ int.jl:49
   %4 = icmp sgt i32 %3, -1
; └
  %5 = select i1 %4, double 0.000000e+00, double 0x4070000010000010
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:170 within `Float32'
; ┌ @ float.jl:60 within `Float64'
   %6 = sitofp i32 %3 to double
; └
; ┌ @ float.jl:410 within `muladd'
   %7 = fmul contract double %6, 0x3E70000010000010
   %8 = fadd contract double %7, %5
; └
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:170 within `Float32' @ float.jl:251
  %9 = fptrunc double %8 to float
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:170 within `Float32'
  ret float %9
}
```
on Julia 1.0. It is sometimes joked that `@pure` means "ask Jameson if it's OK," but here I think it's pretty straightforward.

src/normed.jl
timholy: You're right that making this more straightforward seems desirable.

timholy: With
```diff
diff --git a/src/normed.jl b/src/normed.jl
index 737657b..07f5b86 100644
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -114,7 +114,7 @@ end
 _exp2(::Val{N}) where {N} = exp2(N)
 
 # for Julia v1.0, which does not fold `div_float` before inlining
-inv_rawone(x) = (@generated) ? (y = 1.0 / rawone(x); :($y)) : 1.0 / rawone(x)
+@pure inv_rawone(x) = 1.0 / rawone(x)
 
 function (::Type{T})(x::Normed) where {T <: AbstractFloat}
     # The following optimization for constant division may cause rounding errors.
```
I get
```julia
julia> x = N8f24(15.2)
15.2N8f24

julia> @code_llvm Float32(x)

;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:159 within `Float32'
define float @julia_Float32_17656(%jl_value_t addrspace(10)*, { i32 } addrspace(11)* nocapture nonnull readonly dereferenceable(4)) {
top:
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:160 within `Float32'
; ┌ @ Base.jl:20 within `getproperty'
   %2 = getelementptr inbounds { i32 }, { i32 } addrspace(11)* %1, i64 0, i32 0
; └
; ┌ @ int.jl:498 within `unsafe_trunc'
; │┌ @ int.jl:464 within `rem'
    %3 = load i32, i32 addrspace(11)* %2, align 4
; └└
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:169 within `Float32'
; ┌ @ promotion.jl:349 within `<' @ int.jl:49
   %4 = icmp sgt i32 %3, -1
; └
  %5 = select i1 %4, double 0.000000e+00, double 0x4070000010000010
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:170 within `Float32'
; ┌ @ float.jl:60 within `Float64'
   %6 = sitofp i32 %3 to double
; └
; ┌ @ float.jl:410 within `muladd'
   %7 = fmul contract double %6, 0x3E70000010000010
   %8 = fadd contract double %7, %5
; └
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:170 within `Float32' @ float.jl:251
  %9 = fptrunc double %8 to float
;  @ /home/tim/.julia/dev/FixedPointNumbers/src/normed.jl:170 within `Float32'
  ret float %9
}
```
on Julia 1.0. It is sometimes joked that `@pure` means "ask Jameson if it's OK," but here I think it's pretty straightforward.

kimikage: To tell the truth, I understand the purpose of `@pure`, but I don't understand the function of `@pure` at all. I never realized the benefits of `@pure`, so I did not even think of ​​using `@pure`. Thanks!

Well, unfortunately, in fact, I never realized the benefits of `@pure`. Because of the *constant folding*, we should not use `code_llvm` or `code_native` here.
```julia
julia> x = N8f24(15.2)
15.2N8f24

julia> @code_typed Float32(x)
CodeInfo(
159 1 ─       goto #3 if not false                                                               │
    2 ─       nothing::Nothing                                                                   │
160 3 ─ %3  = (Base.getfield)(x, :i)::UInt32                                                     │╻       getproperty
    │   %4  = (Base.bitcast)(Int32, %3)::Int32                                                   ││╻       rem
161 └──       goto #5 if not false                                                               │
    4 ─       nothing::Nothing                                                                   │
164 5 ─       goto #7 if not false                                                               │
    6 ─       nothing::Nothing                                                                   │
169 7 ─ %9  = (Base.slt_int)(%4, 0)::Bool                                                        │╻       <
    │         (Base.ifelse)(true, 0x00ffffff, 0x00000000)::UInt32                                ││╻╷╷╷╷╷  rawone
    │   %11 = (Base.uitofp)(Float64, 0x00ffffff)::Float64                                        │││╻╷╷╷    promote
    │   %12 = (Base.div_float)(1.0, %11)::Float64                                                │││╻       /
    │   %13 = (Base.mul_float)(4.29497e9, %12)::Float64                                          │╻       *
    │   %14 = (FixedPointNumbers.ifelse)(%9, %13, 0.0)::Float64                                  │
170 │   %15 = (Base.sitofp)(Float64, %4)::Float64                                                │╻       Type
    │         (Base.ifelse)(true, 0x00ffffff, 0x00000000)::UInt32                                ││╻╷╷╷╷╷  rawone
    │   %17 = (Base.uitofp)(Float64, 0x00ffffff)::Float64                                        │││╻╷╷╷    promote
    │   %18 = (Base.div_float)(1.0, %17)::Float64                                                │││╻       /
    │   %19 = (Base.muladd_float)(%15, %18, %14)::Float64                                        │╻       muladd
    │   %20 = (Base.fptrunc)(Base.Float32, %19)::Float32                                         │╻       Type
    └──       return %20                                                                         │
) => Float32
```
Do you have any other suggestions? :sob:

timholy: If it's correct in the LLVM IR, nothing else matters.

kimikage: Could you go into more detail?
This is a matter of the speed, not the *correctness*. 
https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-553185335
https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-554698883

kimikage: Could you go into more detail?
This is a matter of the speed, not the *correctness*. 
https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-553185335
https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-554698883

timholy: Oh, I wasn't looking at vectorization, sorry. The point is that the LLVM is the last stage before native code.

If `@pure` doesn't work, let's keep your `@generated` version.

kimikage: > The point is that the LLVM is the last stage before native code.

I get the point, but even if it's correct in the LLVM IR...

> Especially, in cases of `Normed{UInt32,f}`->`Float32` where `2 <= f <= 24`, the LLVM generates a destructive conditional branch.

```llvm
  %5 = select i1 %4, double 0.000000e+00, double 0x4070000010000010
```
-> :boom: :boom: :boom:
```nasm
        vxorpd  %xmm0, %xmm0, %xmm0
        testl   %eax, %eax
        jns     L30
        vmovsd  .LCPI639_0, %xmm0       # xmm0 = mem[0],zero
L30:
```
I think there is almost no way to intervene in the LLVM backend from Julia scripts. By using `-0.0` instead of `0.0`, the conditional branch (`jns`) can be removed, but it stalls in the SIMD version. So I gave up!

timholy: I probably don't know as much about SIMD as you do, but is `select` actually destructive? It's not technically a branch. See https://github.com/eschnett/SIMD.jl#representing-simd-vector-types-in-julia (search for `select`) and http://blog.llvm.org/2011/12/llvm-31-vector-changes.html (suggesting it went into LLVM 3.1).

Of course, whether it applies these is another question.

I didn't mean to dissuade you from the `@generated` solution, by all means adopt that if it works better. I personally would also not stress too much about Julia 1.0 performance, if it's really painful. "Use a more recent Julia version" is a perfectly fine response to a performance concern.

kimikage: In general, `select` is not destructive. Since this is a *rare* (technically interesting but personally hateful) case, I just left the comment. This is a problem with the *non-SIMD* LLVM IR as you dumped above. Moreover, although I don't like *the* branch, it is still reasonable in points of the binary size (i.e. cache efficiency) and the high-performance branch prediction of modern CPUs. (This usecase unfortunately does not go well with the branch prediction, but the LLVM backend cannot know that.)

> I didn't mean to dissuade you from the `@generated` solution, by all means adopt that if it works better.

I am aware of that. It is just my hesitation. The following comment which I wrote above:

> BTW, I am thinking of moving these low-level utility functions and macros to a new file (e.g. "src/utilities.jl")

implicitly suggests that perhaps I will add more `@generated` (e.g. `div_rawones(a, b) = rawone(a) / rawone(b)`) without the users noticing. I do not think it is pretty good to (over-)use `@generated`, but it is certainly powerful.

Now, I am not going to resist the temptation. This commit is *far* from what I originally envisioned, but I am happy because `float(3N5f3) == 3` comes `true`! :satisfied:

PR #141, opened 2019-11-22
Remove the obsolete promotions for reductions (Fixes #137)

### LINKED ISSUE #137: Removing the obsolete promotions for reductions
The current codebase has no compatibility with Julia v0.6, because the syntax for parametric methods has been *fully* changed to the current (i.e.`where`) style. I think the backporting is not easy. Therefore, the following codes are no longer useful.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/70ae1d6ed27c7005191f23b56343e98ff9ef2026/src/FixedPointNumbers.jl#L157-L169

Removing the obsolete promotions may help improve the promotion rules. For example, the current promotions use `const Treduce = Float64`, but I don't think it is always pretty good. (I can expect the negative impact of the braking change to be significant, though.)
Moreover, it *ostensibly* improves the code coverage.

kimikage:
This fixes #137.

FYI, v0.4.6 is the last release which officially supports Julia v0.6.
https://github.com/JuliaMath/FixedPointNumbers.jl/releases/tag/v0.4.6

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -2,8 +2,6 @@ VERSION < v"0.7.0-beta2.199" && __precompile__()
 
 module FixedPointNumbers
 
-using Base: reducedim_initarray
-
 import Base: ==, <, <=, -, +, *, /, ~, isapprox,
              convert, promote_rule, show, isinteger, abs, decompose,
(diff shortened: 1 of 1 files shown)


timholy:
Thanks!

Issue #41, opened 2016-03-30
Overflow checked arithmetics


phlavenk:
As discussed in [https://github.com/JuliaLang/julia/issues/15690](url) with @timholy I've been recently bitten by the inconsistency regarding overflow in FixedPoint arithmetic when dealing with Images.jl.

 FixedPointNumbers are pretending to be `<:Real` numbers, but their behavior regarding arithmetic is more like `Integer`.  I just cannot think of any use-case one would get any advantage from the "modulo" arithmetic when dealing with the numbers guaranteed to fall within <0, 1> range. I'd be glad if my code stopped by an OverflowError exception indicating problem in my algorithm. Thus, easy fixable by manual widening of the operands. With the current silent overflow, algorithm just continues, giving finally wrong results.

Before writing a PR to introduce arithmetic that throws on overflow or underflow, I'd like to know your opinion on this change. I was also thinking of using a global "flag" to dynamically opt-in for the overflow-checking behavior but i'm worried about the branching costs. Is it possible e.g. use traits for this behavior change?

timholy:
Agreed that this can be an annoyance, but the alternative is annoying, too. Operations with FixedPointNumbers used to return `Float32` or `Float64` (see #12), and that changed for consistency with base julia in #27.

Let's consider alternatives. With images, one approach should be `float(img1)+img2`. `float` was deprecated for Base, but it's too useful in the image processing world to not have it. Another (more performant) option would be to make the output type the first argument, `+(Image{Float32}, img1, img2)` instead of `img1+img2`. Both of these have strengths and weaknesses. Any other ideas?

phlavenk:
The problem is, usually you don't even know you should take care of the overflow arithmetics when dealing with images (recall `FixedPoint <: Real`). Usually, any math (applying kernels, multiplying with a number etc.) you do to a pixel does the widening implicitly (just because you don't specify the constant as UFixed8): 
```
println(typeof(im[1,1]))
println(typeof(im[1,1]*1))

```
gives you implicit widening
```
ColorTypes.Gray{FixedPointNumbers.UFixed{UInt8,8}}
ColorTypes.Gray{Float64}
```
Unfortunately overflow pops in quickly as soon as you try to subtract pixelvalue from an other pixelvalue (both UFixed)  and one forgets to convert the operands to a float. 

[...]

Issue #126, opened 2019-06-16
mixing Fixed and Normed


rapus95:
I recently had to add several differences of `N0f8` values to finally take the `abs` of it. Thus I tried to do
```julia
#a,b of type N0f8
x = zero(Q7f8)
x += a - b
```
which probably would've failed due to the difference being negative and thus not representable.
Thus the next step for me would've been to convert `a,b` to `Q7f8` which didn't work.

Thus my question: Is there any reason why we can't mix up these types?
btw `promote(N0f8(.5), Q7f8(.5))` throws an error.

johnnychen94:
An interesting result (perhaps a bug) related to this is:

```julia
julia> promote_type(UInt8, Int8)
UInt8

julia> UInt8(1) - Int8(2)
0xff
```

timholy:
We should define promotions that make sense. Care to give it a crack?

In practice, the two number types are slightly different: `N1f7` is scaled by 1/127, but `Q0f7` is scaled by `1/128`. What this means is that arithmetic may be faster if you first convert everything to `Float32`, and then store the final result in whatever format you want.

Issue #139, opened 2019-11-18
Commonizing code between `Fixed` and `Normed`


kimikage:
`Fixed` and `Normed` have evolved independently. Therefore, some functions are specialized even though they do not need specialization.

cf. ~https://github.com/JuliaMath/FixedPointNumbers.jl/compare/ee5bd547bf73cf9ec6b976088c0c2565630f4ea5...kimikage:commonize~ PR #151 

I'm going to refactor the codes by taking the following 3 steps.

1. Commonize the existing implementation code
2. Add the functions which are implemented only in either (e.g. rounding functions for `Fixed`)
3. Commonize the test codes

Although "test-first" is a good practice, the step 3. requires major renovations. Please give me your advice if you have any ideas.

kimikage:
It might be a good idea to add the step.
0. Update comments and Add docstrings

For example, "32-bit fixed point" is a holdover from the *`Fixed32` age*.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/8d177395f327abe8cf9725b81a3b302fab9c856b/src/fixed.jl#L1-L2

If we modify the comments, I think it is better to write them in the docstring style, e.g:
```julia
"""
    FixedPoint{T <: Integer, f} <: Real

Supertype of the two fixed-point number types: `Fixed{T, f}` and `Normed{T, f}`.

The parameter `T` is the underlying machine representation and `f` is the number of fraction bits.
"""
abstract type FixedPoint{T <: Integer, f} <: Real end
```

The problem is that I am not good at English.

kimikage:
Steps 1 and 2 are almost complete. Step 0 (the coding guidelines) is partially unfinished. Step 3 has not been started yet. I will keep this issue open until we have a concrete work plan.

kimikage:
I thought the test code was difficult to commonize. However, as I'll add tests for arithmetic operations, I am not very satisfied with the current test code.

A practical problem is the long testing times. It's not fatal at the moment, but it should be an obstacle to supporting signed `Normed` in the future. 
We spend  the most part of the testing time compiling. Therefore, except the reduction of memory allocation, optimization techniques for "run-time" speed has almost no effect. So it's important to reduce the number of test cases, or target types.
Currently, 128-bit/64-bit types are tested, even though they are rarely used in practice. I believe that the reduction of the target type can be abstracted.

kimikage:
[...]

PR #143, opened 2019-11-27
WIP/RFC/Scary: allow signed Normed numbers

timholy:
This is a partial implementation of a speculative change, allowing signed variants of `Normed`. For example, `S7f8` would represent the range `-128.502` to `128.498`; like `N0f8` it is still scaled by 255, but also has 1 sign bit and 7 bits to store absolute values bigger than 1.

The motivation here is to support differences, related to #126, #41, and coming up with a good accumulator type in [CheckedArithmetic](https://github.com/JuliaMath/CheckedArithmetic.jl).

Speculatively, we could use `S7f8` for subtraction (and maybe addition) with `N0f8`. Pros: computing differences among pixels in images is a common operation, and it would be nice for this to be less of a gotcha than it is currently. Cons: this is against convention in the Julia ecosystem, and in opposition to #27. Thoughts? I'd especially like to hear from @vchuravy in addition to other recent contributors.

### DIFF
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -1,22 +1,34 @@
-# Normed{T,f} maps UInts from 0 to 2^f-1 to the range [0.0, 1.0]
+# Normed{T,f} maps Integers from 0 to 2^f-1 to the range [0.0, 1.0]
 # For example, Normed{UInt8,8} == N0f8 maps 0x00 to 0.0 and 0xff to 1.0
 
-struct Normed{T<:Unsigned,f} <: FixedPoint{T,f}
+struct Normed{T<:Integer,f} <: FixedPoint{T,f}
     i::T
 
-    Normed{T, f}(i::Integer,_) where {T,f} = new{T, f}(i%T)   # for setting by raw representation
+    function Normed{T, f}(i::Integer,_) where {T,f}    # 2-arg form for setting by raw representation
+        isa(f, Int) || error("f must be an Int")
+        0 < f <= sizeof(T)*8 - (T<:Signed) || error("f must be between 1 and the number of non-sign bits")
+        new{T, f}(i%T)
+    end
 end
 
 Normed{T, f}(x::AbstractChar) where {T,f} = throw(ArgumentError("Normed cannot be constructed from a Char"))
 Normed{T, f}(x::Complex) where {T,f} = Normed{T, f}(convert(real(typeof(x)), x))
 Normed{T, f}(x::Base.TwicePrecision) where {T,f} = Normed{T, f}(convert(Float64, x))
-Normed{T1,f}(x::Normed{T2,f}) where {T1 <: Unsigned,T2 <: Unsigned,f} = Normed{T1,f}(convert(T1, x.i), 0)
 
-typechar(::Type{X}) where {X <: Normed} = 'N'
-signbits(::Type{X}) where {X <: Normed} = 0
+Normed{T1,f}(x::Normed{T2,f}) where {T1 <: Integer,T2 <: Integer,f} = Normed{T1,f}(convert(T1, x.i), 0)
 
-for T in (UInt8, UInt16, UInt32, UInt64)
-    for f in 1:sizeof(T)*8
+typechar(::Type{X}) where X <: Normed{T} where T<:Unsigned = 'N'
+typechar(::Type{X}) where X <: Normed{T} where T<:Signed   = 'S'
+signbits(::Type{X}) where X <: Normed{T} where T<:Unsigned = 0
+signbits(::Type{X}) where X <: Normed{T} where T<:Signed   = 1
+
+I64(::Type{<:Unsigned}) = UInt64
+I64(::Type{<:Signed})   = Int64
+I32(::Type{<:Unsigned}) = UInt32
+I32(::Type{<:Signed})   = Int32
+
+for T in (UInt8, UInt16, UInt32, UInt64, Int16, Int32, Int64)
+    for f in 1:sizeof(T)*8-(T<:Signed)
         sym = Symbol(String(take!(showtype(_iotypealias, Normed{T,f}))))
         @eval begin
             const $sym = Normed{$T,$f}
@@ -25,11 +37,11 @@ for T in (UInt8, UInt16, UInt32, UInt64)
     end
 end
 
-reinterpret(::Type{Normed{T,f}}, x::T) where {T <: Unsigned,f} = Normed{T,f}(x, 0)
+reinterpret(::Type{Normed{T,f}}, x::T) where {T <: Integer,f} = Normed{T,f}(x, 0)
 
-zero(::Type{Normed{T,f}}) where {T,f} = Normed{T,f}(zero(T),0)
-function oneunit(::Type{T}) where {T <: Normed}
-    T(typemax(rawtype(T)) >> (8*sizeof(T)-nbitsfrac(T)), 0)
+zero(::Type{Normed{T,f}}) where {T <: Integer,f} = Normed{T,f}(zero(T),0)
+function oneunit(::Type{N}) where N <: Normed{T,f} where {T,f}
+    N(typemax(rawtype(N)) >> (8*sizeof(N)-nbitsfrac(N)-(T<:Signed)), 0)
 end
 one(::Type{T}) where {T <: Normed} = oneunit(T)
 zero(x::Normed) = zero(typeof(x))
@@ -37,65 +49,70 @@ oneunit(x::Normed) =  one(typeof(x))
 one(x::Normed) = oneunit(x)
 rawone(v) = reinterpret(one(v))
 
-# Conversions
-function Normed{T,f}(x::Normed{T2}) where {T <: Unsigned,T2 <: Unsigned,f}
+# More construction and conversion
+function Normed{T,f}(x::Normed{T2}) where {T <: Integer,T2 <: Integer,f}
     U = Normed{T,f}
     y = round((rawone(U)/rawone(x))*reinterpret(x))
-    (0 <= y) & (y <= typemax(T)) || throw_converterror(U, x)
+    (typemin(T) <= y) & (y <= typemax(T)) || throw_converterror(U, x)
     reinterpret(U, _unsafe_trunc(T, y))
 end
 N0f16(x::N0f8) = reinterpret(N0f16, convert(UInt16, 0x0101*reinterpret(x)))
 
-(::Type{U})(x::Real) where {U <: Normed} = _convert(U, x)
+(::Type{N})(x::Real) where {N <: Normed} = _convert(N, x)
 
-function _convert(::Type{U}, x) where {T, f, U <: Normed{T,f}}
-    if T == UInt128 # for UInt128, we can't widen
+function _convert(::Type{N}, x) where {T <: Integer, f, N <: Normed{T,f}}
+    if T === UInt128 || T === Int128 # for [U]Int128, we can't widen
         # the upper limit is not exact
-        (0 <= x) & (x <= (typemax(T)/rawone(U))) || throw_converterror(U, x)
-        y = round(rawone(U)*x)
+        ((typemin(T)/rawone(N)) <= x) & (x <= (typemax(T)/rawone(N))) || throw_converterror(N, x)
+        y = round(rawone(N)*x)
     else
-        y = round(widen1(rawone(U))*x)
-        (0 <= y) & (y <= typemax(T)) || throw_converterror(U, x)
+        y = round(widen1(rawone(N))*x)
+        (typemin(T) <= y) & (y <= typemax(T)) || throw_converterror(N, x)
     end
-    reinterpret(U, _unsafe_trunc(T, y))
+    reinterpret(N, _unsafe_trunc(T, y))
 end
 # Prevent overflow (https://discourse.julialang.org/t/saving-greater-than-8-bit-images/6057)
-function _convert(::Type{U}, x::Float16) where {T, f, U <: Normed{T,f}}
-    if Float16(typemax(T)/rawone(U)) > Float32(typemax(T)/rawone(U))
-        x == Float16(typemax(T)/rawone(U)) && return typemax(U)
+function _convert(::Type{N}, x::Float16) where {T <: Integer, f, N <: Normed{T,f}}
+    if Float16(typemax(T)/rawone(N)) > Float32(typemax(T)/rawone(N))
+        x == Float16(typemax(T)/rawone(N)) && return typemax(N)
+        if T <: Signed
+            x == Float16(typemin(T)/rawone(N)) && return typemin(N)
+        end
     end
-    return _convert(U, Float32(x))
+    return _convert(N, Float32(x))
 end
-function _convert(::Type{U}, x::Tf) where {T, f, U <: Normed{T,f}, Tf <: Union{Float32, Float64}}
-    if T == UInt128 && f == 53
-        0 <= x <= Tf(3.777893186295717e22) || throw_converterror(U, x)
+function _convert(::Type{N}, x::Tf) where {T <: Integer, f, N <: Normed{T,f}, Tf <: Union{Float32, Float64}}
+    if T === UInt128 && f == 53
+        Tf(0) <= x <= Tf(3.777893186295717e22) || throw_converterror(N, x)
+    elseif T === Int128 && f == 53
+        Tf(-1.888946593147859e22) <= x <= Tf(1.888946593147859e22) || throw_converterror(N, x)
     else
-        0 <= x <= Tf((typemax(T)-rawone(U))/rawone(U)+1) || throw_converterror(U, x)
+        Tf((typemin(T)+rawone(N))/rawone(N)-1) <= x <= Tf((typemax(T)-rawone(N))/rawone(N)+1) || throw_converterror(N, x)
     end
 
-    significand_bits = Tf == Float64 ? 52 : 23
+    significand_bits = Tf === Float64 ? 52 : 23
     if f <= (significand_bits + 1) && sizeof(T) * 8 < significand_bits
-        return reinterpret(U, unsafe_trunc(T, round(rawone(U) * x)))
+        return reinterpret(N, unsafe_trunc(T, round(rawone(N) * x)))
     end
     # cf. the implementation of `frexp`
     Tw = f < sizeof(T) * 8 ? T : widen1(T)
     bits = sizeof(Tw) * 8 - 1
-    xu = reinterpret(Tf == Float64 ? UInt64 : UInt32, x)
+    xu = reinterpret(Tf === Float64 ? I64(T) : I32(T), x)
     k = Int(xu >> significand_bits)
-    k == 0 && return zero(U) # neglect subnormal numbers
+    k == 0 && return zero(N) # neglect subnormal numbers
     significand = xu | (one(xu) << significand_bits)
     yh = unsafe_trunc(Tw, significand) << (bits - significand_bits)
     exponent_bias = Tf == Float64 ? 1023 : 127
     ex = exponent_bias - k + bits - f
     yi = bits >= f ? yh - (yh >> f) : yh
     if ex <= 0
-        ex == 0 && return reinterpret(U, unsafe_trunc(T, yi))
-        ex != -1 || signbit(signed(yi)) && return typemax(U)
-        return reinterpret(U, unsafe_trunc(T, yi + yi))
+        ex == 0 && return reinterpret(N, unsafe_trunc(T, yi))
+        ex != -1 || signbit(signed(yi)) && return typemax(N)
+        return reinterpret(N, unsafe_trunc(T, yi + yi))
     end
-    ex > bits && return reinterpret(U, ex == bits + 1 ? one(T) : zero(T))
+    ex > bits && return reinterpret(N, ex == bits + 1 ? one(T) : zero(T))
     yi += one(Tw)<<((ex - 1) & bits) # RoundNearestTiesUp
-    return reinterpret(U, unsafe_trunc(T, yi >> (ex & bits)))
+    return reinterpret(N, unsafe_trunc(T, yi >> (ex & bits)))
 end
 
 rem(x::T, ::Type{T}) where {T <: Normed} = x
@@ -125,14 +142,18 @@ function (::Type{T})(x::Normed) where {T <: AbstractFloat}
     convert(T, y)  # needed for types like Float16 which promote arithmetic to Float32
 end
 
-function Base.Float16(x::Normed{Ti,f}) where {Ti <: Union{UInt8, UInt16, UInt32}, f}
+# A slightly faster copysign (one that avoids type-piracy)
+setsign(x::Float32, i::UInt32) = x
+setsign(x::Float32, i::Int32)  = reinterpret(Float32, reinterpret(UInt32, x) | (reinterpret(UInt32, i) & reinterpret(UInt32, typemin(Int32))))
+
+function Base.Float16(x::Normed{Ti,f}) where {Ti <: Union{UInt8, UInt16, UInt32, Int8, Int16, Int32}, f}
     f == 1 ? Float16(x.i) : Float16(Float32(x))
 end
-function Base.Float16(x::Normed{Ti,f}) where {Ti <: Union{UInt64, UInt128}, f}
+function Base.Float16(x::Normed{Ti,f}) where {Ti <: Union{UInt64, UInt128, Int64, Int128}, f}
     f == 1 ? Float16(x.i) : Float16(Float64(x))
 end
 
-function Base.Float32(x::Normed{UInt8,f}) where f
+function Base.Float32(x::Normed{<:Union{UInt8,Int8},f}) where f
     f == 1 && return Float32(x.i)
     f == 2 && return Float32(Int32(x.i) * 0x101) * @f32(0x550055p-32)
     f == 3 && return Float32(Int32(x.i) * 0x00b) * @f32(0xd4c77bp-30)
@@ -143,7 +164,7 @@ function Base.Float32(x::Normed{UInt8,f}) where f
     f == 8 && return Float32(Int32(x.i) * 0x155) * @f32(0xc0f0fdp-40)
     0.0f0
 end
-function Base.Float32(x::Normed{UInt16,f}) where f
+function Base.Float32(x::Normed{<:Union{UInt16,Int16},f}) where f
     f32 = Float32(x.i)
     f ==  1 && return f32
     f ==  2 && return f32 * @f32(0x55p-8)  + f32 * @f32(0x555555p-32)
@@ -155,19 +176,19 @@ function Base.Float32(x::Normed{UInt16,f}) where f
     f == 16 && return f32 * @f32(0x01p-16) + f32 * @f32(0x010001p-48)
     Float32(x.i / rawone(x))
 end
-function Base.Float32(x::Normed{UInt32,f}) where f
+function Base.Float32(x::Normed{T,f}) where {T <: Union{UInt32,Int32}, f}
     f == 1 && return Float32(x.i)
     i32 = unsafe_trunc(Int32, x.i)
     if f == 32
         rh, rl = Float32(i32>>>16), Float32((i32&0xFFFF)<<8 | (i32>>>24))
-        return muladd(rh, @f32(0x1p-16), rl * @f32(0x1p-40))
+        return setsign(muladd(rh, @f32(0x1p-16), rl * @f32(0x1p-40)), x.i)
     elseif f >= 25
         rh, rl = Float32(i32>>>16),Float32(((i32&0xFFFF)<<14) + (i32>>>(f-14)))
-        return muladd(rh, Float32(@exp2(16-f)), rl * Float32(@exp2(-14-f)))
+        return setsign(muladd(rh, Float32(@exp2(16-f)), rl * Float32(@exp2(-14-f))), x.i)
     end
     # FIXME: avoid the branch in native x86_64 (non-SIMD) codes
     m = ifelse(i32 < 0, 0x1p32 * inv_rawone(x), 0.0)
-    Float32(muladd(Float64(i32), inv_rawone(x), m))
+    return setsign(Float32(muladd(Float64(i32), inv_rawone(x), m)), x.i)
 end
 function Base.Float32(x::Normed{Ti,f}) where {Ti <: Union{UInt64, UInt128}, f}
     f == 1 ? Float32(x.i) : Float32(Float64(x))
@@ -176,7 +197,11 @@ end
 function Base.Float64(x::Normed{Ti,f}) where {Ti <: Union{UInt8, UInt16}, f}
     Float64(Normed{UInt32,f}(x))
 end
-function Base.Float64(x::Normed{UInt32,f}) where f
+function Base.Float64(x::Normed{Ti,f}) where {Ti <: Union{Int8, Int16}, f}
+    Float64(Normed{Int32,f}(x))
+end
+
+function Base.Float64(x::Normed{<:Union{UInt32,Int32},f}) where f
     f64 = Float64(x.i)
     f ==  1 && return f64
     f ==  2 && return (f64 * 0x040001) * 0x15555000015555p-72
@@ -206,7 +231,7 @@ function Base.Float64(x::Normed{UInt32,f}) where f
     f == 32 && return (f64 * 0x010101) * 0x00ff0000ffff01p-96
     f64 / rawone(x)
 end
-function Base.Float64(x::Normed{UInt64,f}) where f
+function Base.Float64(x::Normed{UInt64,f}) where f   # TODO: optimized version for Int64
     f == 1 && return Float64(x.i)
     if f >= 53
         rh = Float64(unsafe_trunc(Int64, x.i >> 16)) * @exp2(16-f) # upper 48 bits
@@ -215,7 +240,7 @@ function Base.Float64(x::Normed{UInt64,f}) where f
     end
     x.i / rawone(x)
 end
-function Base.Float64(x::Normed{UInt128,f}) where f
+function Base.Float64(x::Normed{UInt128,f}) where f   # TODO: optimized version for Int128
     f == 1 && return Float64(x.i)
     ih, il = unsafe_trunc(Int64, x.i>>64), unsafe_trunc(Int64, x.i)
     rh = Float64(ih>>>16) * @exp2(f <= 53 ? 80 : 80 - f) # upper 48 bits
@@ -241,7 +266,8 @@ Base.Rational{Ti}(x::Normed) where {Ti <: Integer} = convert(Ti, reinterpret(x))
 Base.Rational(x::Normed) = reinterpret(x)//rawone(x)
 
 # Traits
-abs(x::Normed) = x
+abs(x::Normed{<:Unsigned}) = x
+abs(x::T) where T<:Normed  = T(abs(x.i), 0)
 
 (-)(x::T) where {T <: Normed} = T(-reinterpret(x), 0)
 (~)(x::T) where {T <: Normed} = T(~reinterpret(x), 0)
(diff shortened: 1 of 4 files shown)


kimikage:
Wow, this is a great challenge!
I agree with the aim and concept, but this is a dramatic change, you know. So, I cannot immediately determine whether the benefits of generalization exceed those of specialization.

From a *practical* point of view, since `N0f8` cannot be converted to *8*-bit signed `Normed`, I don't think it has much merit. For addition / subtraction, `N8f8` can be used instead of `S7f8`, and `Float32` is suitable for many other operations.
Of course, I'll do all I can for this challenge.

timholy:
> From a practical point of view, since N0f8 cannot be converted to 8-bit signed Normed, I don't think it has much merit... For addition / subtraction, N8f8 can be used instead of S7f8

Not entirely. I don't entirely subscribe to what follows, but for the purposes of argument let me make the extreme case. Consider
```
julia> 0x01 - 0x02
0xff
```
If magnitude matters, it's bad when `smaller - small = big`. This is much more serious a problem than `0xff + 0x01 == 0x00`, because it's much more important to be able to accurately measure numbers close to 0 than it is to `typemax(T)`. (We often use small integers on a 64-bit machine, but rarely do we use integers close to ±2^63.) A very common operation is to look at the difference between images, e.g., to detect changes in a scene, and the default arithmetic rules here make this sufficiently fraught that newbies are almost guaranteed to write buggy code. Somewhat along the lines of https://stackoverflow.com/questions/10168079/why-is-size-t-unsigned, one could declare that one might ban unsigned types from arithmetic altogether.

With this line of thinking, there are two options: (1) deliberately lose one bit of precision, and for an 8-bit image demote it to 7-bit and then use the extra bit as a sign bit (so that 8-bit images get loaded as `S0f7`); (2) for any arithmetic operation involving unsigned types, immediately go to the next wider signed type. In option (2), when doing any mathematics at all, we promote `Nmfn` (where `m+n` is 8, 16, or 32) to `S(2m+n-1)fn` so that we get a sign bit with no loss of precision; as a bonus we can also store larger values, but that's not really the main goal. Note that arithmetic with `Spfq` (where `q + q` is 7, 15, or 31) does *not* promote, consistent with the rest of Julia. This is just an unsigned-types-are-bad-news decision, not a decision to widen all arithmetic operations.

Again, I'm making the "extreme" case to show why this is worth considering. A more balanced view would point out two important negatives:

- it's different from the rest of Julia
- when applied to arrays, it causes the memory requirement to double

The latter is a problem, but the fact that it's limited to a doubling (and not, say, 8x) and that RAM is cheap means that I don't think the latter is fatal. I think the former is the more serious disadvantage of such a change.

If you ask me, what is my honest opinion, I am not sure. I support the fact that operations with `UInt8` return `UInt8`. I am less sure that I support operations with `N0f8` returning `N0f8`; to me it seems that no one chooses `UInt8` without thinking about it, but for image processing I bet most users don't choose an element type at all, they take whatever arrives when you load the image from disk.

timholy:
We should make sure that major contributors to/watchers of JuliaImages are aware of this discussion: CC @kmsquire, @rsrock, @SimonDanisch, @Evizero, @tejus-gupta, @annimesh2809, @zygmuntszpak, @RalphAS, @Tokazama, @juliohm, @bjarthur, @ianshmean, @jw3126, @tlnagy. Sorry for the spam (and sorry if I omitted someone who might want to know).

johnnychen94:
I'm skeptical of its practical usage.

> I am less sure that I support operations with `N0f8` returning `N0f8`

unless all operations are done under `N0f8` without conversion, which is impossible as far as I can see, this feature is of less practical usage. A more common situation at present is:

```julia
# not returning N0f8 for N0f8 input
img_in = load(imgfile) # N0f8
img_temp = foo1(img_in) # N0f8->Float32
img_out = foo2(img_temp) # Float32
img_out = n0f8(img_out) # Float32->N0f8
```

Note that there're only two conversions.

But if it became:
```julia
# returning N0f8 for N0f8 input
img_in = load(imgfile) # N0f8
img_temp = foo1(img_in) # N0f8->Float32->N0f8
img_out = foo2(img_temp) # N0f8->Float32->N0f8
```

There're four now, where two of them are unnecessary. If it's about saving memory usage, `Float16` might be a better choice (although the current performance is said poor).

I once opened an issue related to what I said here (preserving type): https://github.com/JuliaImages/ImageCore.jl/issues/82.

vchuravy:
It's been a while since I deeply thought about FixedPointNumbers and the difference between Fixed and Normed.

> Cons: this is against convention in the Julia ecosystem, and in opposition to #27.

I see there are two orthogonal concerns.

1. Is a number normed or fixed point. This is a subtle concern that tripped me up in the past when trying to unify these two concepts. Both have fractional bits, but the meaning differs slightly.

2. Is there a sign bit. 

From a representational perspective I don't think there is any strong reason to not have signed normed numbers, although looking at my comment history I once remarked that I had no longer a strong reason for having unsigned fixed point numbers.

The normed numbers always had an odd place in the ecosystem, since I see them primarily as storage numbers and not as computational numbers.
Then there is the question of semantics, which normed numbers mostly derive from what is needed in the Color ecosystem and as far as I can tell if
one does computation on them, one might want to do saturating computation or one needs to change representation. But the semantics we have for Normed is overflow/underflow and that makes the least sense in image calculations. 

In conclusion I don't have strong opinions and only existential questions. It might make sense to divorce fixed and normed and make normed an explicit color representational/computational type instead of having to worry about general semantics.

timholy:
From @johnnychen94 :

> A more common situation at present is:

> ```julia
> # not returning N0f8 for N0f8 input
> img_in = load(imgfile) # N0f8
> img_temp = foo1(img_in) # N0f8->Float32
> ```

and @vchuravy:

> The normed numbers always had an odd place in the ecosystem, since I see them primarily as storage numbers and not as computational numbers.

Yep. Both of these express the notion that the current values are mostly storage and not computation. If you already think this way, there's little advantage to this proposal.

However, we do support computation with such values. Since I think I may not have been sufficiently clear, let me walk you through an interesting experiment. This is one that someone might do to compare two images, or look for frame-by-frame temporal differences in a movie:
```julia
julia> using ImageView, TestImages

julia> mri = testimage("mri");

julia> diff1 = mri[:,:,1:end-1] - mri[:,:,2:end];

julia> diff2 = float.(mri[:,:,1:end-1]) - float.(mri[:,:,2:end]);

julia> imshow(diff1); imshow(diff2)
```
It's worth doing this on your own, but to make life easy here are the first frames from each:
d1
d2

The second one is "correct," with mid-gray encoding no difference. Or you can use `colorsigned` from `ImageCore`:
d4
d3

(magenta is positive and green is negative)

Now, if you're used to calling `float(img)` before doing any computations, there is nothing to worry about. But new users often are bothered by this kind of overflow (e.g., @johnnychen94, you got bit by this in https://github.com/JuliaImages/ImageCore.jl/issues/63).

To the specifics of this proposal: aside from defining & supporting the type (which, as @kimikage indicated, is a burden), the change that would "fix" these issues is a single line, which could be written:

```julia
-(x::N0f8, y::N0f8) = S7f8(Int16(x.i) - Int16(y.i), 0)
```

This is definitely a breaking change, but it's not a huge computational burden:
```julia
julia> minus1(x, y) = x - y   # this is essentially what we do now
minus1 (generic function with 1 method)

julia> minus2(x, y) = Int16(x) - Int16(y)    # this is what we could do if we had S7f8
minus2 (generic function with 1 method)
```
when passed `UInt8`s is just the difference in "padding" with a byte of all zeros:
```julia
julia> @btime minus1(x, y) setup=(x = rand(UInt8); y = rand(UInt8))
  1.240 ns (0 allocations: 0 bytes)
0x22

julia> @btime minus2(x, y) setup=(x = rand(UInt8); y = rand(UInt8))
  1.239 ns (0 allocations: 0 bytes)
187
```
So in contrast to what @johnnychen94 said above, there isn't much/any of a conversion overhead to worry about. (For comparison, `float(x)` is more expensive, clocking in at 2ns.)

Now, just about the only operation this would affect are + and -; as soon as you multiply by a `AbstractFloat` you convert to floats and the rest of the computations would not be affected.

So, to recap, the costs of this are:

- more breaking changes
- more types to support
- a doubling in size of images if you add or subtract them

and the benefits are

- adding and subtracting will more often work more in the way users expect them to.

That's pretty much the bottom line, I think.

johnnychen94:
I'm satisfied with the "breaking change" and it's more like bug fixes to me.

An alternative way to fix this issue is making `Normed` a pure storage type and promotes to `float` or `Fixed` for every relevant operation. That could really save a lot of effort.

kimikage:
> So in contrast to what @johnnychen94 said above, there isn't much/any of a conversion overhead to worry about. (For comparison, float(x) is more expensive, clocking in at 2ns.)

As I showed in the past benchmarks, the benchmarks of a *single* operation can mislead us.

src/normed.jl
```diff
@@ -25,77 +37,82 @@ for T in (UInt8, UInt16, UInt32, UInt64)
...
+            x == Float16(typemin(T)/rawone(N)) && return typemin(N)
+        end
     end
-    return _convert(U, Float32(x))
+    return _convert(N, Float32(x))
 end
-function _convert(::Type{U}, x::Tf) where {T, f, U <: Normed{T,f}, Tf <: Union{Float32, Float64}}
-    if T == UInt128 && f == 53
-        0 <= x <= Tf(3.777893186295717e22) || throw_converterror(U, x)
+function _convert(::Type{N}, x::Tf) where {T <: Integer, f, N <: Normed{T,f}, Tf <: Union{Float32, Float64}}
```
kimikage: I'm sorry but this method is currently optimized for non-negative numbers and cannot handle negative numbers. (pay attention to `k`.) Of course, this is a trivial issue, but I mean, "optimization is specialization".

src/normed.jl
```diff
@@ -241,7 +266,8 @@ Base.Rational{Ti}(x::Normed) where {Ti <: Integer} = convert(Ti, reinterpret(x))
 Base.Rational(x::Normed) = reinterpret(x)//rawone(x)
 
 # Traits
-abs(x::Normed) = x
+abs(x::Normed{<:Unsigned}) = x
+abs(x::T) where T<:Normed  = T(abs(x.i), 0)
```
kimikage: Why does not `abs(::Normed{<:Signed})` return `Normed{<:Unsigned}`? (Of course, this is not a question.)

kimikage:
It might be better to clarify what we need.
- Extra bits
- Sign bit (i.e. negative numbers)
- Newbie-friendly promotion
- Speed

It is premature to discuss the speed because the current `Fixed` and `Normed` are not fully optimized.
If we just need the extra bits for subtractions, we can use `N8f8` instead of `S7f8`, although it is never newbie-friendly.
With regards to the image processing, it is important that there are no negative RGB colors. Thus, we should (implicitly) apply the mapping.

vchuravy:
> I also agree about that. So, I think there are two types of the type hierarchy:

Yeah I have a strong preference for the first one since it emphasises the semantic difference between Normed and Fixed.

kimikage:
Since we can only choose one for now, I agree completely. However, I think choosing the former is not denying the latter. If we explicitly choose the former, we should be responsible for the implicit nature of the latter. `Normed{<:Signed}` breaks the nature of the latter.

src/normed.jl
timholy: Try this: `abs(Int8(-5))`. In general Julia doesn't change the number type, even to the point where `abs(typemin(Int8))` returns a negative number!

src/normed.jl
timholy: Yeah, I hadn't checked any of this, this is WIP.

timholy:
> With regards to the image processing, it is important that there are no negative RGB colors. Thus, we should (implicitly) apply the mapping.

Not sure I fully agree here. We allow folks to construct out-of-gamut colors routinely. This is important in, e.g., computing the mean color of an image patch, where the natural implementation of mean is `sum(a)/length(a)`. It's nice to have a type that enforces in-gamut, but that's not essential. Similar considerations apply in https://www.aristeia.com/Papers/C++ReportColumns/sep95.pdf (linked from the C++ link above).

> Normed{<:Signed} is not good, as mentioned above.

I wasn't sure which place, can you elaborate? I don't think anything in the concept of "normed" enforces "unsignedness," so I don't see any problem with `Normed{<:Signed}`. You can always check for `Signed` in the type paramter if you need to differentiate (which I needed to do in a couple of places in this PR, though did not check in enough!).

src/normed.jl
kimikage: I don't care about that much *in `FixedPointNumbers`*. However, it is generally not good to ask other packages for such verification and modification. If this is *essential* to the design, we should change it aggressively, but I don't think so.

src/normed.jl
kimikage: That's right. Therefore, it can be concluded that `-(x :: N0f8, y :: N0f8) = S7f8 (Int16 (x.i)-Int16 (y.i), 0)` is not good! :stuck_out_tongue_closed_eyes:

src/normed.jl
timholy: Not sure I understand. All I meant was that this was a draft I decided to let people know I was working on. I will fix this before we merge this if we decide to do so.

src/normed.jl
timholy: I agree, this is in conflict with the rest of Julia. I guess we could return `Normed{<:Unsigned}`, since the only other "always agrees with expectations" is to widen which also changes the type.

src/normed.jl
kimikage: I think this is a simple problem. The *current* `Normed` is always unsigned, but your `Normed` is not always unsigned.

kimikage:
I'm sorry. I'm not good with words.

> We allow folks to construct out-of-gamut colors routinely.

What I want to say is not about the *gamut*, but about the visualization. Since our LCDs  cannot represent negative RGB colors, we need to project abstract numbers into concrete colors. It usually involves the type conversion. If you do type conversion after all, is there any reason not to convert the numbers to a more convenient type (e.g. `Float32`, `Float16`, `Fixed`) first? If you do not visualize the data, what is the difference between `N8f8` and `S7f8` in RAMs?

Although signed `Normed`s are useful as an intermediate storage type, I don't realize the need to implement those arithmetic.

timholy:
Good! This focuses the discussion down. Really I think we have 3 options:

1. The current approach: implement math with Normed numbers, with the results also Normed. **Advantage**: consistent with the rest of Julia. **Disadvantage**: "wrong" results are returned for very simple operations.
2. Declare that Normed numbers are storage only. Math always returns floating-point equivalents,  e.g., `+(x::N0f8, y::N0f8) = float(x) + float(y)`. **Advantages**: (1) "correct" results are always returned, although roundoff will become more annoying, and (2) considerable simplification of this package and the type landscape. **Disadvantages**: (1) less consistent with the rest of Julia. (2) there's a 4x blowup of size for most operations. (`Float16` is too slow to be a practical result type.) (3) For some algorithms, like [median filtering](https://en.wikipedia.org/wiki/Median_filter) and [histogram equalization](https://en.wikipedia.org/wiki/Histogram_equalization), there are particularly efficient implementations that rely on the discretization provided by 8-bit values.
3. Implement signed variants of Normed numbers. **Advantages**: "correct" results are usually returned (barring overflow), and roundoff won't happen since the underlying types are integral. **Disadvantages**: (1) less consistent with the rest of Julia, (2) more types we have to support, (3) like option 2, discretization becomes less of an advantage even by moving to 16-bit types (there are more bins you have to check).

timholy:
In comparing 2 vs 3, perhaps the key question is whether the extra complexity is worth optimizing a small number of operations (basically, addition and subtraction). For reference,

```julia
julia> function mysum(acc, A)
           @inbounds @simd for a in A
               acc += a
           end
           return acc
       end
mysum (generic function with 1 method)

julia> A = collect(rand(N0f8, 100));

julia> using BenchmarkTools

julia> Base.:(+)(x::N56f8, y::N0f8) = N56f8(x.i + y.i, 0)

julia> @btime mysum(zero(N56f8), A)
  23.733 ns (1 allocation: 16 bytes)
50.255N56f8

julia> @btime mysum(0.0f0, A)
  23.544 ns (1 allocation: 16 bytes)
50.254906f0

julia> @btime mysum(zero(N56f8), A)
  22.852 ns (1 allocation: 16 bytes)
50.255N56f8

julia> @btime mysum(0.0f0, A)
  23.782 ns (1 allocation: 16 bytes)
50.254906f0
```
so at least on my machine the performance difference seems to be within the noise. (I am surprised, TBH.)

Can others try this benchmark and report back?

kimikage:
These trends apply not only to x86_64 but also to ARM (for NEON).

```julia
julia> versioninfo()
Julia Version 1.0.3
Platform Info:
  WORD_SIZE: 32
  LIBM: libopenlibm

julia> @btime mysum(zero(N56f8), A)
  1.219 μs (2 allocations: 32 bytes)
47.776N56f8

julia> @btime mysum(0.0f0, A)
  1.698 μs (1 allocation: 8 bytes)
47.776474f0
```

timholy:
OK, so there are architectures where it makes a 2x difference. Is that (along with its domain of applicability, which is basically summation/differencing) enough to matter?

kimikage:
I think this is a compiler issue rather than a CPU architecture or CPU performance issue. Of course, compiler-friendly types are better. However, is that really what you wanted to know about the benchmark?

kimikage:
There are two orthogonal concepts for accumulation.
- Using `Float` for the accumulator
- Converting each element to `Float`
```julia
julia> function mysum_raw(acc, A)
           @inbounds @simd for a in A
               acc += reinterpret(a)
           end
           return acc / FixedPointNumbers.rawone(eltype(A))
       end
```
This is just a conceptual code and not practical. This difference might be noticeable for large (more practical) arrays.

timholy:
I don't think most algorithms should dive into this level of detail about their element types. Your code above *should* be equivalent to `mysum(zero(N56f8), A)` (on a 64-bit machine) in all respects except the result type, and of course it can be converted to `Float64` depending on what you do with it next. Relying on Julia's abstraction properties means we can write generic algorithms that don't have to be hand-tuned for each element type, without paying a performance penalty. But what we decide here will end up determining what that inner loop looks like.

kimikage:
My conceptual code is never never never practically good. Is it related to this discussion? :confused:
I don't want to explain the problems of your benchmark code in detail

timholy:
> I mainly support the option 1, but I do not object to adding isolated signed Normed numbers.
I am strongly against the option 2. If you do that so, you should split FixedPointNumbers into FixedPointTypes and FixedPointArithmetic.

>  Is it related to this discussion?

https://github.com/JuliaMath/FixedPointNumbers.jl/pull/143#issuecomment-559965758 is, in my mind (flaws included), a benchmark for whether 3 provides any advantage over 2. Given that you don't like option 2, and that at least on some architectures 3 does perform better than 2, it seems our main options are 1 and 3. Option 3 would allow one to fix #41, #126, https://github.com/JuliaImages/ImageCore.jl/issues/63. Option 1 is the status quo.

Can we agree on 3 as a reasonable consensus solution?

kimikage:
First of all, a wrong benchmark misleads us. Since `sum` is a special case of addition, it is not fair. Moreover, the optimizer still has room for improvement. Do you want to "redesign" every time the compiler changes?

I think the main problem of #41 (and https://github.com/JuliaImages/ImageCore.jl/issues/63 ?) will be solved by `@checked` in the near future. I think #126 can be solved without Option 3.

What are the specific problems which Option 1 cannot solve?

timholy:
> will be solved by `@checked` in the near future

I don't doubt you already know everything I'm about to say, so I may not be understanding your true question. But let me try: `@checked` supports "normal" arithmetic with the extra feature that it throws an error if the results overflow. For any nontrivial image pair, `@checked img1 - img2` would result in an error. Is that a solution? I guess it depends on your perspective. Would one always want `@checked` arithmetic? Doubtful, since image-processing is performance-sensitive. So what is `@checked` for? Mainly tests and debugging, I think. It only works on surface syntax; to go "deeper" we'd need a separate package, `CheckedFixedPointNumbers` that checks all operations by default. I am fairly certain that there isn't a good mechanism that (1) lets you have just one type, (2) works with precompilation, (3) allows you to switch between checked and unchecked arthmetic, and (4) doesn't also incur a performance penalty. And again, the only thing that buys you is an error when overflow is detected.

Let me try to get others to chime in by making the options here crystal clear:

1. we can have `img1 - img2` work for arbitrary images but use "wraparound" behavior. This is the status quo.
2. we can have `img1 - img2` throw a MethodError, forcing users to write `float.(img1) - float.(img2)`. This will avoid the "wraparound" behavior.
3. we can have `img1 - img2` promote eltypes automatically to something that doesn't typically wrap around. This comes in two flavors:
  a. promote to `float` for all operations
  b. promote to `Normed{<:Signed}` for addition and subtraction

If I understand the voting so far, it looks like this (:+1: means approve, :-1: means disapprove, nothing means relatively neutral):

|               | Option 1: status quo | Option 2: force conversion | Option 3a: promotion to float | Option 3b: promotion to `Normed{<:Signed}` |
|:-------------:|:--------------------:|:--------------------------:|:-------------------:|:-------------------:|
|   @kimikage   |           :+1:           |                           |          :-1:           | :-1: |
| @johnnychen94 |           :-1:          |              :+1:              |                    |  |
|    @timholy   |           :+1:          |              :-1:             |          :+1:           |  :+1: |

If I've misunderstood people's votes, please correct me. How do others vote?

kimikage:
Your explanation is very easy to understand. :+1:

Option 2:
force conversion: :+1: 
delete the code supporting arithmetic: :-1:

I don't know much about the costs of education, but generally it's not bad to be able to manage your own workspace yourself.

Whether the answer is correct or not is determined by the specifications. `FixedPointNumbers` has no way of knowing the user's specifications. Wait for someone to create `FormalMethods.jl`. However, the higher the *level*, the closer to the users.

In the first place, where do `Normed` numbers come from in image processing? Do the users really *create* them? I think breaking or narrowing the *entry pathway* is a common tactics. I can't agree with this much, but this is better than Option 3.

timholy:
To clarify, option 2 = "force conversion" is equivalent to "delete the code supporting arithmetic." The only way to force users to convert to some other type before doing arithmetic is for arithmetic on `Normed` types to throw an error. Options 2 is the "Normed types are for storage only" option. So your votes on Option 2 are in conflict with one another. If you don't want to delete the code supporting arithmetic, you have to decide what type arithmetic returns, leaving options 1, 3a, and 3b.

> where do Normed numbers come from in image processing

`img = load("mypicture.png")`. They are what we return for any integer-encoded image (e.g., 8- and 16-bit images).

I've checked every package in JuliaRegistries/General that depends on FixedPointNumbers. The only one that *isn't* related to image processing or colors is https://github.com/JuliaAudio/SampledSignals.jl, which from their README would apparently benefit from `Normed{<:Signed}`.

johnnychen94:
Option 2: neutral 😑 
I'm only unsatisfied with the status quo. Option 2 is what users of FixedPointNumbers.jl can do to solve this issue at present. I can see the importance of Option 3(especially Option 3b), but since I'm not familiar with the implementation details and not care much about it at present, really I have nothing much to comment on.

> What are the specific problems which Option 1 cannot solve?

If it's on test stage, `ReferenceTest.jl` already bring us a tool to examine the numerical results on different types. The problem with Option 1 is that we can't guarantee the numerical accuracy _without test_.

Plug-and-play is a very common style when we're processing images. As a playground, we generally don't write tests. When wrap around behavior happens, we would tend to conclude that "my processing pipeline should be modified" and it's actually quite severe mislead in many research scenarios.

> Where do Normed numbers come from in image processing? Do the users really create them?

I only manually convert Float32 to Normed when I need to benchmark different algorithms -- take advantages of the inaccuracy of representation -- just to have a more stable benchmark result, e.g., 

```julia
julia> float_x = 0.1
0.1

julia> float_y = 0.103
0.103

julia> N0f8(float_x) - N0f8(float_y)
0.0N0f8
```

timholy:
@johnnychen94, thanks for clarifying; I've updated your votes in the table accordingly.

One thing I should also make sure is understood: Python's `scikit-image` uses Option 1; Matlab uses Option 1 but uses saturating arithmetic. If we choose something different, we're blazing a different path. But we're already doing that with FixedPointNumbers anyway; other frameworks that support both "fixed point" (integer) images and floating-point images are content to live with the equation `255 == 1.0 (sometimes)`, whereas we are not. Moreover, both Matlab & Python have to make this decision in the context of general operations on arrays with a very limited palette of element types, whereas we're in the situation of having an image-specific element type and hence have greater freedom to set our own rules. Consequently, I don't think it's at all crazy to do something "better" than what they do, if indeed it is possible to do anything that is better.

Consequently I understand @johnnychen94's dissatisfaction with the current status quo, even if I also acknowledge that it's one of the most self-consistent options we have.

kimikage:
@timholy:
> To clarify, option 2 = "force conversion" is equivalent to "delete the code supporting arithmetic."

I understand and clearly disagree with Option 2.

> The only way to force users to convert to some other type before doing arithmetic is for arithmetic on `Normed` types to throw an error.

I cannot completely agree with that. I just proposed another way. (The way is not pretty good, though.)

> I've checked every package in JuliaRegistries/General that depends on FixedPointNumbers.

Great survey! However, *isolated* signed `Normed` and `Normed{<:Signed}` should be clearly distinguished. The former should not be detrimental to the users (except for the unnatural naming), but the latter is a breaking change, even though it may be a minor change for you. Again, "optimization is specialization". 

@johnnychen94:
> The problem with Option 1 is that we can't guarantee the numerical accuracy *without test*.

That's right! Therefore, I think testing is very important. A program with no specifications or tests is a *poem*. Although I do never hate poetry, it is not the basis for software design.
If the errors mislead us, we should consider throwing *correct* and useful errors first. I don't think this is an obligation of *low-level* packages like `FixedPointNumbers`.  Of course, it is good to improve `FixedPointNumbers` for the better errors.

----

This is just a thought experiment. What about `NormedFloat` instead of `Normed{<:Signed}`? I can assure you that the fully-optimized `NormedFloat` is not inferior to `Normed{<:Signed}` in terms of speed and size. However, I don't think it is good. What is the advantage of `Normed{<:Signed}` over `NormedFloat`?

timholy:
> This is just a thought experiment. What about NormedFloat instead of Normed{<:Signed}? I can assure you that the fully-optimized NormedFloat is not inferior to Normed{<:Signed} in terms of speed and size. However, I don't think it is good. What is the advantage of Normed{<:Signed} over NormedFloat?

What is NormedFloat?

> How much time do we have for this discussion?

It can go on paralyzing development forever, probably :wink:. Seriously, we need to reach some kind of consensus here. Why don't you propose what you think we should do, rather than just argue against what I am proposing? AFAICT the only solution currently is the status quo. I can live with that but I'm not happy with it. Perhaps you feel you have proposed a solution, but if so I didn't understand it.

timholy:
> I understand and clearly disagree with Option 2.

I see you edited your post to clarify. Yes, it's fine to throw an informative message. I would have done it that way, I was being a bit casual by saying we'd just delete them. Delete the *content*, I should have specified.

kimikage:
> What is NormedFloat?

I'm sorry for the lack of explanation, but it is so simple that no explanation is needed. `NormedFloat` means `Normed{T<:AbstractFloat, f}`. (Of course, it is no longer a part of `FixedPointNumbers`. So, I did not call it `Normed`.) As a matter of fact, I have already proposed [this concept](https://github.com/JuliaMath/FixedPointNumbers.jl/pull/143#issuecomment-560036636).
We can use `Normed{Float16, 8}` instead of `Normed{Int16, 8}`. 
Regarding the addition and subtraction of `N0f8` numbers (i.e. "safe" operations), `Normed{Float16,8}` is *empirically estimated* to be ~2x slower than (unoptimized) `S7f8`. If the difference of speed is like that, I *empirically assume* that the bottlenecks are memory allocations and memory accesses, and that the practical speeds will be approximately the same. Of course, the sizes are the same.
Well, this is just a thought experiment. I don't think the promotion to `Normed{<:AbstractFloat}` is good, too.

kimikage:
> Perhaps you feel you have proposed a solution, but if so I didn't understand it.

I don't think it is nice, but narrowing the *entry pathway* is one solution, isn't it?

```
julia> img = load("mypicture.png")
┌ Warning: Please specify the pixel type, e.g, `load(Float32, "mypicture.png")`. You are using `N0f8` now.
└ @ Foo
2×2 Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}:
 RGB{N0f8}(0.867,0.157,0.169)  RGB{N0f8}(0.984,0.133,0.851)
 RGB{N0f8}(0.855,0.984,0.945)  RGB{N0f8}(0.553,0.549,0.133)
```

Again, I don't think this is nice. However, I'm not against this because this is the approach which is closer to the users than the low-level `FixedPointNumbers`.

kimikage:
My biggest concern is that there are very few people participating in this discussion rather than what the final decision is.

If you are convinced that signed `Normed` is great, why not complete the implementation of signed `Normed` without breaking changes? If it is really great, everyone will extend the current `Normed`. And then, there should be a movement to promote to signed `Normed`.
Of course, it might be too late for the funding.

I do not intend to inhibit it maliciously. However, I think we should do #139 and supporting `@checked` first. Unfortunately, the development in parallel should cause the conflicts.

ianshmean:
Just FYI, I have been following this, but I’ve not felt like my thoughts are yet robust enough to contribute. 

Perhaps I’d just say a few things:
- silent wraparound keeps me up at night, so I do like the idea of dealing with this
- I love the idea of finding a type promotion system that maintains minimal lossless precision throughout operations (3b), but figuring out the types of each possible combination sounds like a big undertaking, and it would be interesting to see what the benefits in performance and data integrity were over float. 

I guess it seems like some more fleshed-out working case studies could help frame the discussion?

HTH!

timholy:
> NormedFloat

I can't see any advantage in NormedFloat: why have an implicit `/255` in a type that can represent the result directly? The main purpose of these types is to harmonize the values of "integer-valued" and "floating-point valued" images, which are treated differently in most image-processing frameworks (e.g., https://scikit-image.org/docs/stable/user_guide/data_types.html). This package harmonizes them by, for example, providing a number type with values ranging from 0 to 1 that's representable in 8 bits. I don't see how `NormedFloat` contributes to that, since you can just do the division directly.

> but narrowing the entry pathway is one solution to "force conversion"

We can't force conversion of a 1TB image. My lab uses those all the time, via memory-mapping.

However, I suppose we could change all the image-loading routines to return a [MappedArray](https://github.com/JuliaArrays/MappedArrays.jl) using `of_eltype`. That would lazily-convert to Float32 representation upon usage. That would make the rules of arithmetic in this package essentially irrelevant. Thoughts?

> If you are convinced that signed Normed is great, why not complete the implementation of signed Normed without breaking changes?

Yes, we could create the type and then decide how to use it later. I just don't feel like creating types for the fun of it, I'd rather be sure there's a path towards them being useful, hence the proposal to change the rules of arithmetic.

Breaking changes aren't a disaster if they are managed carefully; we can bound all packages that use FixedPointNumbers to 0.x versions, and then the packages that want to keep up can install CompatHelper to learn about the new approach. Breaking changes are acceptable especially for a 1.0 release; we broke just about everything in the Julia 0.6->1.0 transition. We already have some breaking changes queued, though any change in arithmetic rules would be a much larger change. 

So for me the point of this discussion was to determine whether we can find a better way. I am thinking we can't, so perhaps I should abandon this direction.

timholy:
@kimikage, I may have been remiss in not directing you to this earlier, but if you haven't read it yet I highly recommend reading https://juliaimages.org/latest/arrays_colors/, particularly starting at https://juliaimages.org/latest/arrays_colors/#fixedpoint-1 and reading to the end of the page. It's worth comparing to that scikit-image page I linked above, particularly section titled "Input types" which, you'll note, forces one to change the value when you want to change the representation. I apologize for the lack of a Japanese translation, I am aware that it makes it much more difficult. I would *not* fare well if the only documentation were available in Japanese :man_shrugging: .

kimikage:
In the first place, please don't put the blame on my lack of knowledge. It is my personal problem that I am stubborn, so you can blame just me any amount. However, my (or *our*) claim is that we should keep it consistent with the rest of Julia. No matter how much I learn, the Julia language will never change unless I take action.

tlnagy:
> However, I suppose we could change all the image-loading routines to return a MappedArray using `of_eltype`. That would lazily-convert to Float32 representation upon usage. That would make the rules of arithmetic in this package essentially irrelevant. Thoughts?

This change would fix the problem, but we would lose lossless precision doing so and it would increase the cost of accessing a view of the image since you would need to do this conversion every time.

After reading through this thread, here are my thoughts:

- Option 1: :-1: I know first-hand that this isn't very user-friendly for new users and is well discussed earlier in this thread. That said, it's gotten us this far, but it's problematic.
- Option 2: :-1: This is very user-unfriendly even if we provide an explanation for what to do. Things _should_ just work, even 3a is preferable to this. 
- Option 3a: :-1: Eager: Not realistic for large images* Lazily: See my comments above
- Option 3b: :+1: I quite like this solution. I think that if the benchmarks for realistic image sets are good, then I would prefer this solution since it would balance the tradeoffs between the different options.

* I think I've had to resort to Option 3a (recasting images to float) in my own code, which I'm not happy with (ref: https://github.com/tlnagy/SegmentationTools.jl/blob/36d6b98236cf62880e171059547fd55f9f11e9b4/src/utils.jl#L39). 

> Consequently, I don't think it's at all crazy to do something "better" than what they do, if indeed it is possible to do anything that is better.

I would go further and say that we would be doing the greater imaging community a disfavor if we didn't try and develop something better given all the problems with the status quo and that we are in a unique position to build it.

> So for me the point of this discussion was to determine whether we can find a better way. I am thinking we can't, so perhaps I should abandon this direction.

I hope not! I like the bones here and the status quo could be better and changes like these are why I love the JuliaImages ecosystem :)

johnnychen94:
FYI, I used a lot of `MappedArrays.of_eltype` to do the lazy conversion in `ImageDistances.jl`, and the routine I take there is:

1. `N0f8` -> `Float32`
2. `*(::Float32, ::Float32)`
3. `Float32` -> `N0f8`

kimikage:
> I can't see any advantage in NormedFloat

`NormedFloat` *must not* be good because it is a spoiling candidate or a touchstone. Please tell *us* the advantages of `Normed{<:Signed}`, not the disadvantages of `NormedFloat`. At least, `NormedFloat` is the *same* as`Normed{<:Signed}` in the following ways:
> Implement signed variants of Normed numbers. Advantages: "correct" results are usually returned (barring overflow), and roundoff won't happen since the underlying types are integral. Disadvantages: (1) less consistent with the rest of Julia, (2) more types we have to support, (3) like option 2, discretization becomes less of an advantage even by moving to 16-bit types (there are more bins you have to check).


> Breaking changes aren't a disaster if they are managed carefully;

Ideally. Please face the facts(#139). We are only interested in what we need. I hate *potential* minor bugs. They are more vicious than obvious malfunctions.

> I don't care about that much *in `FixedPointNumbers`*. However, it is generally not good to ask other packages for such verification and modification. If this is *essential* to the design, we should change it aggressively, but I don't think so.

----
Please give me some time to study. My claims are already written, so it's okay to conclude during my absence.

kimikage:
I do not mean to be rude, but the supporters of "3b" *had better* clarify the promotion rules. (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/issues/147#issuecomment-561670150)
`Normed` has many types other than `N0f8` and `N0f16`, even though they are not often used in image processing.

Tokazama:
Been following this but don't have the bandwidth right now to help think upon solutions too much. The only thing I really care about on this is that:

1. Type conversion doesn't became a nightmare. I'm sure a motivated individual could figure out a complex set of promotion rules given a bunch of types, but that's always obnoxious.
2. It doesn't affect speed in downstream plotting functions.

I think both of these have been addressed to some degree, so the only other thing I'd add is that I only use fixed point numbers as a storage type passed to graphics libraries. Most of the time it doesn't make sense for an MRI to be directly represented as a fixed point number.

PR #145, opened 2019-11-29
Fix performance regression with inlining failure on Julia v1.3.0 (Fixes #144)

### LINKED ISSUE #144: Performance regression in `Normed` -> `Float` conversions  on Julia v1.3.0
I have confirmed that Julia v1.2.0 and v1.3.0 give almost similar results on `Normed`->`Float` conversions (#129, #138). However, I found the performance regression (~2x - 3x slower) on x84_64 machines in the following cases:
- `Vec4{N0f32}` -> `Vec4{Float32}`
- `Vec4{N0f64}` -> `Vec4{Float32}`
- `Vec4{N0f64}` -> `Vec4{Float64}`

(cf. https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-544244496)

I'm *not* going to rush to investigate the cause or fix this problem. I  submit this issue as a placeholder in case any useful information is found.

kimikage:
Julia 1.3.0 generates more redundant intermediate codes which are eventually optimized.
This trivial change reduces the redundant intermediate codes to promote inlining. (Fixes #144)

### DIFF
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -159,10 +159,10 @@ function Base.Float32(x::Normed{UInt32,f}) where f
     f == 1 && return Float32(x.i)
     i32 = unsafe_trunc(Int32, x.i)
     if f == 32
-        rh, rl = Float32(i32>>>16), Float32((i32&0xFFFF)<<8 | (i32>>>24))
+        rh, rl = Float32(i32>>>0x10), Float32((i32&0xFFFF)<<0x8 | i32>>>0x18)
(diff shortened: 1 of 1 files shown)


vchuravy:
Do you want to resurrect #69 ;) would be good to have measurements for before and after

kimikage:
I personally like the idea of ​​#69. However, the result is like a horoscope and is not really useful for *most* people.:sweat_smile:

Issue #157, opened 2019-12-28
Problems with conversions from `Rational`


kimikage:
`Normed` supports the conversion to `Rational`, but does not support the conversion from `Rational`.
```julia
julia> N0f8(1//2)
ERROR: MethodError: no method matching unsafe_trunc(::Type{UInt8}, ::Rational{Int64})
```
----

`Fixed` supports the conversion from `Rational`, but it overflows too easily.
```julia
julia> Q0f7(1//2)
ERROR: InexactError: trunc(Int8, 128)
```
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7ad0f0c460961024191f672332fcf3d7cef655d1/src/fixed.jl#L15

Issue #154, opened 2019-12-25
Inconsistent `Integer` results


kimikage:
The return types of  `Integer(::Fixed)` and `Integer(::Normed)` are based on different manners.

#### `Fixed{T}`: `T` 
```julia
julia> typeof(Integer(1Q1f6))
Int8

julia> typeof(Integer(1Q1f14))
Int16
```
#### `Normed`: `Int` with (non-intuitive) promotions
```julia
julia> typeof(Integer(1N0f8))
Int64

julia> typeof(Integer(1N0f16))
Int64

julia> typeof(Integer(1N0f32))
Int64

julia> typeof(Integer(1N0f64)) # !?
BigInt
```
#### `Float`: `Int` w/o promotions
```julia
julia> typeof(Integer(Float16(1)))
Int64

julia> typeof(Integer(1.0f0))
Int64

julia> typeof(Integer(1.0))
Int64

julia> typeof(Integer(0x1p63))
ERROR: InexactError: Int64(9.223372036854776e18)
```
I think it is better to unify the manners, even though it will be a breaking change. 
I consider it appropriate for `FixedPoint` to return numbers in the rawtype `T` as `Integer(::Fixed)` returns.

FYI, with fix for issue #153, we can use `floor(Integer, x)` etc.

kimikage:
BTW, there is no longer a zero-argument `InexactError()`.
```julia
> julia> Integer(1.5Q1f6)
ERROR: MethodError: no method matching InexactError()
Closest candidates are:
  InexactError(::Symbol, ::Any, ::Any) at boot.jl:280
```
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7ad0f0c460961024191f672332fcf3d7cef655d1/src/fixed.jl#L67-L74
Also, I don't think this specialization makes much sense.

kimikage:
There is the problem of type inconsistency in `Rational`, too.

```julia
julia> typeof(Rational(0.5Q1f6))
Rational{Int64}

julia> typeof(Rational(0.5Q0f7))
Rational{Int64}

julia> typeof(Rational(0.5Q1f62))
Rational{Int128}

julia> typeof(Rational(0.5Q0f63))
Rational{Int128}
```

```julia
julia> typeof(Rational(0.5N0f8))
Rational{UInt8}

julia> typeof(Rational(0.5N0f64))
Rational{UInt64}
```
[...]

PR #151, opened 2019-12-23
Commonize code between `Fixed` and `Normed`

kimikage:
This PR is the first part of #139.
I am very sorry about a conflict with #143, but in the long-term, I think this PR will help enhance the functions (e.g. checked arithmetic).

Because there are many changes, it is recommended that you track changes on a per-commit basis.

This PR includes a minor breaking change, i.e., this change the exception type of `one(Q0f7)` etc. from `InexactError` to `ArgumentError`.

I'm ready to fix `isinteger` (issue #120, PR #123) after the merging this PR. And then, I'll fix the issue #150.

**Edit:**
Rescheduling: this -> **issue #153** -> `isinteger` -> `rem` -> **issue #157**

Issue #150, opened 2019-12-15
Rounding errors in `rem` of `Normed`


kimikage:
As I mentioned in #102, PR #131 did not change the `rem`, so this is a reminder.

> To avoid impact on the performance, this PR does not change the `rem`. 
> https://github.com/JuliaMath/FixedPointNumbers.jl/blob/da3931838dd3a5f03e8ed00542a306d685002e33/src/normed.jl#L67-L70
> 
> Therefore, this does not fix the problems such as:
> ```julia
> julia> using Colors
> 
> julia> white = parse(RGB{Float32}, "white")
> RGB{Float32}(1.0f0,1.0f0,1.0f0)
> 
> julia> convert(RGB{N0f32}, white)
> RGB{N0f32}(0.0,0.0,0.0)
> ```

_Originally posted by @kimikage in https://github.com/JuliaMath/FixedPointNumbers.jl/pull/131#issuecomment-547238807_

```julia
julia> using FixedPointNumbers

julia> for f = 1:32
           F = Normed{UInt32,f}
           rem_one = 1.0f0 % F
           rem_one == one(F) || @show rem_one
       end
rem_one = 1.00000003N7f25
rem_one = 1.00000001N6f26
rem_one = 1.000000007N5f27
rem_one = 1.000000004N4f28
rem_one = 1.000000002N3f29
rem_one = 1.0000000009N2f30
rem_one = 1.0000000005N1f31
rem_one = 0.0N0f32
```

kimikage:
Perhaps the following is more accurate than the current implementation, and less accurate than PR #131.
```julia
function Base.rem(x::Float32, ::Type{T}) where {f, T <: Normed{UInt32,f}}
    f <= 24 && return reinterpret(T, unsafe_trunc(UInt32, round(rawone(T)*x)))
    r = unsafe_trunc(UInt32, round(Int32, x * Float32(0x1p24)))
    reinterpret(T, r << UInt8(f - 24) - r >> 0x18)
end
```

kimikage:
BTW, I reported this issue because I noticed  that the documents of  Colors.jl include the example with `N0f32`.
https://github.com/JuliaGraphics/Colors.jl/issues/324
http://juliagraphics.github.io/Colors.jl/latest/namedcolors/

I think it would be easier to fix the bug than add a note to the document.

timholy:
[...]

PR #158, opened 2019-12-30
Add rounding functions for `Fixed` (Fixes #153)

### LINKED ISSUE #153: Adding rounding functions for `Fixed`
As I mentioned in #139, there are no rounding functions for `Fixed`.
```julia
julia> round(0.502N0f8)
1.0N0f8

julia> round(0.500Q0f7)
ERROR: MethodError: no method matching round(::Fixed{Int8,7}, ::RoundingMode{:Nearest})
```
An important difference between `Normed` and `Fixed` is that `Normed` has no *tie* (or half) numbers, but `Fixed` has.
So, I also think it is a good idea to support the 2nd argument (i.e. `RoundingMode`) explicitly.

kimikage:
This is a part of #151 and fixes #153.

succeeding PRs: this -> `isinteger`(#120) -> `rem`(#150) -> `Integer`(#154) -> `Rational`(#157)

**Edit:**
~Let's decide #159 first.~ Done.

### DIFF
--- a/src/fixed.jl
+++ b/src/fixed.jl
@@ -50,6 +50,9 @@ function rawone(::Type{Fixed{T,f}}) where {T, f}
     oneunit(T) << f
 end
 
+intmask(::Fixed{T,f}) where {T, f} = -oneunit(T) << f # Signed
+fracmask(x::Fixed{T,f}) where {T, f} = ~intmask(x) # Signed
+
 # unchecked arithmetic
 
 # with truncation:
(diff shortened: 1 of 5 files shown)


timholy:
Let's merge this before #163

kimikage:
Though I merged PR #159, this PR supports `Fixed{Int8,8}` and so on, i.e. I reduced the test cases for `f == 8sizeof(T)` and added TODO comments.
Due to inlining, the `depwarn` is somewhat noisy. (Sorry, I overlooked this.)
In addition, I added support for specifying `RoundingMode` for `round`.

src/fixed.jl
```diff
@@ -91,6 +94,73 @@ end
...
+    f == bitwidth(T) - 1 && return x.i == typemin(T) ? x : zero(x)
+    t = x.i & intmask(x)
+    r = x.i & fracmask(x)
+    _rawone = oneunit(T) << f
+    reinterpret(Fixed{T,f}, (x.i < 0) & (r != 0) ? t + _rawone : t)
+end
+function floor(x::Fixed{T,f}) where {T, f}
+    f == bitwidth(T) && x.i < 0 && throw_converterror(Fixed{T,f}, -1) # TODO: remove this line
+    Fixed{T,f}(x.i & intmask(x), 0)
+end
```
kimikage: I would like to confirm just in case, your implementation is much more complex.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/ebc5efff51c271f2697579625ef47e19d20c39b2/src/fixed.jl#L92-L96
Am I overlooking something?

timholy: No, you're absolutely right, the simpler one looks correct. :+1:

Issue #155, opened 2019-12-25
Upper limit of the number of fractional bits `f`


kimikage:
In theory, the number of fractional bits `f` can be countlessly large, but what is the actual upper limit?
I assumed that it is `8sizeof(T)-1` for `Fixed{T}`, and `8sizeof(T)` for `Normed{T}`.

However, the larger numbers are also used actually. 
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7ad0f0c460961024191f672332fcf3d7cef655d1/test/fixed.jl#L63

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7ad0f0c460961024191f672332fcf3d7cef655d1/test/fixed.jl#L96

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7ad0f0c460961024191f672332fcf3d7cef655d1/test/fixed.jl#L110

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7ad0f0c460961024191f672332fcf3d7cef655d1/test/fixed.jl#L138

Perhaps some functions do not work properly with large `f`, due to implicit upper limits.

**Edit:**
I intend to support `Fixed{T,f}` where `f == 8sizeof(T)`, i.e. throw no exceptions. However, they have no typealias and will not be tested enough.

kimikage:
I also have some concerns about the type of `f`.

> BTW, I think there is a problem in interoperability with specialization using integer literal parameters.
> ```julia
> julia> Normed{UInt32,Int64(1)} === Normed{UInt32,Int32(1)}
> false
> 
> julia> Normed{UInt32,Int64(1)} == Normed{UInt32,Int32(1)}
> false
> ```
> 
> I don't know how serializers (e.g. JLD2) do (de-)serialization.

_Originally posted by @kimikage in https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-554701260_

I think `f <: Integer` is fine except for the ambiguity associated with omitting constraints, but I think `isa(f, Int)` is problematic as a constraint.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/08c7452754b15b2a214bdb844c8eb5534cb4d09d/src/normed.jl#L8

timholy:
I think those test cases are broken, we should probably change them.

> isa(f, Int) is problematic as a constraint.

Why?

kimikage:
> I think those test cases are broken, we should probably change them.

[...]

PR #159, opened 2020-01-02
[RFC] Add depwarn in constructor of Fixed{T,f} where f == 8sizeof(T)

kimikage:
If we officially deprecate `Fixed{Int8, 8}` and so on, it is better to add a depwarn early. (cf. #155)
However, I also think it is another possible option to continue to support them.

This also adds the domain checks for `f` in the constructors. Because of the dead code elimination, there is no performance regression for valid `f` in most use cases.

### DIFF
--- a/src/fixed.jl
+++ b/src/fixed.jl
@@ -15,7 +15,13 @@ struct Fixed{T <: Signed, f} <: FixedPoint{T, f}
 
     # constructor for manipulating the representation;
     # selected by passing an extra dummy argument
-    Fixed{T, f}(i::Integer, _) where {T,f} = new{T, f}(i % T)
+    function Fixed{T, f}(i::Integer, _) where {T, f}
+        if f == bitwidth(T)
+            Base.depwarn("`Fixed` reserves one bit for the sign. Support for `f=$f` with raw type `T=$T` will be removed in a future release.", :Fixed)
+        end
+        0 <= f <= bitwidth(T) || throw(DomainError(f, "f must be between 0 and $(bitwidth(T)-1) (i.e. the number of non-sign bits of `T=$T`)")) # TODO: change the upper limit
+        new{T, f}(i % T)
+    end
 end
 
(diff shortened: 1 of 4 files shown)


src/fixed.jl
```diff
@@ -15,7 +15,13 @@ struct Fixed{T <: Signed, f} <: FixedPoint{T, f}
 
     # constructor for manipulating the representation;
     # selected by passing an extra dummy argument
-    Fixed{T, f}(i::Integer, _) where {T,f} = new{T, f}(i % T)
+    function Fixed{T, f}(i::Integer, _) where {T, f}
+        if f == bitwidth(T)
+            Base.depwarn("The `f`, which is the same as the number of rawtype bits, will be denied in the future.", :Fixed)
```
timholy:
```suggestion
            Base.depwarn("`Fixed` reserves one bit for the sign. Support for `f=$f` with raw type `T=$T` will be removed in a future release.", :Fixed)
```

To avoid a performance hit from the interpolation you might need the `@noinline` trick, i.e.,

```julia
@noinline dowarn(f, T) = Base.depwarn(...)
f == bitwidth(T) && dowarn(f, T)
```
(Probably not necessary if DCE runs before the allocation pass, I'm not sure of the order so you can check the LLVM code if you want to be sure.)

kimikage:
I changed the messages. 
The interpolation costs about 10% extra time. However, it is just a drop in the bucket, because `depwarn` increases processing time by about 2,000,000%.:sweat_smile:

timholy:
But do you pay it even when it doesn't issue the warning? For example, check the LLVM for `Q0f7(0.2)` with the warning in the source and commented out. As long as it does DCE before the `alloca` pass it should be fine.

kimikage:
"2,000,000%" means DCE works fine. I checked `@code_llvm Q0f7(Int8(25), 0)`.
However, it is not a bad idea to check `@code_llvm Q0f7(0.2)` as a "memorial".

kimikage:
I have to submit some PRs and merge them, so there is still time to release v0.8.0. For now, let's step forward.

PR #160, opened 2020-01-07
Work around JuliaLang/julia#34121

kimikage:
Ref JuliaLang/julia#34121

We have to create `release-0.7` branch to release v0.7.1.

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -206,7 +206,9 @@ end
 rand(::Type{T}) where {T <: FixedPoint} = reinterpret(T, rand(rawtype(T)))
 rand(::Type{T}, sz::Dims) where {T <: FixedPoint} = reinterpret(T, rand(rawtype(T), sz))
 
-include("precompile.jl")
-_precompile_()
+if VERSION >= v"1.1" # work around https://github.com/JuliaLang/julia/issues/34121
+    include("precompile.jl")
+    _precompile_()
+end
 
 end # module


kimikage:
BTW, I do not yet fully understand how this workaround works.

timholy:
Since we don't understand the bug, there's a sense in which I don't understand either. But here's the approximate explanation: when a package is not marked `__precompile__(false)`, then Julia will create a serialized version of the package as a `*.ji` file in your `.julia/compiled` directory. This is essentially a [record of a subset of a fresh Julia process's actions](https://github.com/JuliaLang/julia/blob/8f9dd5d02d0afa675220e77c85f1bd872524239c/base/loading.jl#L1148-L1205)  from the point of opening the toplevel file to the end of module initialization.

Just defining methods does not require that they be specialized for specific types. This is what the `precompile` directive does, and since it runs while under the `--output-incremental=yes` flag, it causes additional `MethodInstance`s to be cached to the `*.ji` file. Thus, by making the function that issues the `precompile` directives conditional on Julia version, we save less stuff to the `*.ji` file for Julia 1.0.x than we do for higher versions of Julia.

johnnychen94:
> BTW, I do not yet fully understand how this workaround works.

This issue is first identified in https://github.com/JuliaImages/juliaimages.github.io/pull/98 (the travis log is missing now) and the cause for that is the upgrade of FileIO v1.2.0 from v1.1.0, which only introduces a precompile operation. In simple words according to my understanding: "since precompilation breaks in Julia v1.0.x, then we don't precompile it."

kimikage:
@johnnychen94,I think your comment is correct. However, if so, we have to stop using *all* `precompile()` on v1.0.x. This is not a joke, but a major concern for me.

Please remember the problem with Images.jl. Images.jl depends not only on FileIO.jl but also on Colors.jl, ColorTypes.jl and FixedPointNumbers.jl.
One possibility is that there are some *extra* trigger conditions. Another possibility is that the trigger conditions have been satisfied and simply the tests do not trigger it.

PR #161, opened 2020-01-07
Work around JuliaLang/julia#34121 and Release 0.7.1

kimikage:
This backports PR #160.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FixedPointNumbers"
 uuid = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
-version = "0.7.0"
+version = "0.7.1"
 
 [compat]
 julia = "1"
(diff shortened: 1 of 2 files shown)


kimikage:
BTW, it might be a good idea to protect the `release-*` branches.

timholy:
Looks like `master` is still on 0.7.0?

kimikage:
I haven't yet rewrote the version number because in FixedPointNumbers.jl, it was a habit to rewrite the version number just before release, not just after release. However, we may want to use v0.8.0 (or v0.7.1) now to clarify the order.

timholy:
In general, I only use release branches if I want to make further releases in an older release cycle. Otherwise I just release from `master`.

kimikage:
We already have the some commits with breaking changes, i.e. v0.7 is already in an old release cycle.

Issue #152, opened 2019-12-24
Add `checked_add` and `checked_sub`


kimikage:
As the first measure against #41, I want to add `checked_add` and `checked_sub` (and `checked_neg`, `checked_abs`) so that we can try [`CheckedArithmetic`](https://github.com/JuliaMath/CheckedArithmetic.jl).

I think it is better to support `checked_mul` at the same time as or after the optimization of current multiplication implements. Adding `checked_mul` implies introducing *unchecked mul*, i.e. breaking changes.

kimikage:
BTW, if checked arithmetic becomes the default and the overflow-safe promotion rules are added in the future, I think that using the `@fastmath` macro to access the successor methods to the current unsafe arithmetic might be a good idea. However, I cannot compromise any more than that.

kimikage:
BTW, the integer types of Rust have `saturating_add()`, `wrapping_sub()` and so on, and their default operators check overflows.

The removal of `@fastmath` (changing to  more specific methods) was proposed (cf. https://github.com/JuliaLang/julia/issues/36246). I think adding `add_fast` and so on for `FixedPoint` seems to be a bad idea, now.

One option is to add the `@saturating` and `@wrapping` macros to CheckedArithmetic (or another new package). I think the new macros also solves the problem that `@checked` is too exhaustive.

kimikage:
>> Also, in the future, the default arithmetic may be changed to the checked arithmetic.
>
>What's your plan for eliminating the performance hit?
>
_Originally posted by @timholy in https://github.com/JuliaMath/FixedPointNumbers.jl/issues/209#issuecomment-671347468_

> It is to give up.:sweat_smile: In any case, we haven't reached the stage of deciding that yet. PR #190 is just a small step.

_Originally posted by @kimikage in https://github.com/JuliaMath/FixedPointNumbers.jl/issues/209#issuecomment-671361857_


However, I am interested in [Cassette.jl](https://github.com/jrevels/Cassette.jl). I think the injection is useful in end user codes. If checked arithmetic is needed only for development/debugging and not for the final code, I think that the default arithmetic can be the current wrapping arithmetic.

It might be also a good idea for CheckedArithmetic.jl to use Cassette.jl. However, in that case, it is better to separate the interface and implementation into separate packages. As you know, I have concerns about the `accumulatortype`. (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/pull/146#issuecomment-560099028)

[...]

Issue #140, opened 2019-11-19
Optimizing `Normed` -> `Normed` conversions


kimikage:
As I suggested [here](https://github.com/JuliaMath/FixedPointNumbers.jl/pull/131#discussion_r341069209), the current `Normed` -> `Normed` conversions are inefficient in some cases.

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/8d177395f327abe8cf9725b81a3b302fab9c856b/src/normed.jl#L41-L46

The current conversion method has two problems:
1. It always checks the input range even if there is no need, and may throw the exception.
2. It always uses floating-point operations even if  there is no need.

The former means that the method is not SIMD-suitable.
Regarding the latter, the conversion between types with the same `f` is already specialized.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/ee5bd547bf73cf9ec6b976088c0c2565630f4ea5/src/normed.jl#L13
There is also the `N0f8`->`N0f16` specialization. (I wonder why.)
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/ee5bd547bf73cf9ec6b976088c0c2565630f4ea5/src/normed.jl#L47

I do not think these are urgent problems. However, the optimization may be useful *in the future* to speed up the accumulation (`reduce`). And I just found a (ugly) [workaround](https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-554357889) for the constant division problem, so I am writing this issue as a memorandum or reminder.

The figures below visualize the cases where the optimization is available.
- The positive (greenish) area means that the conversion is overflow-safe (i.e. there is no need for the range checking).
- The negative (reddish) area means that the conversion is unsafe (i.e. it may throw the exception).
- The deep-colored cells means that the conversion does not need floating-point operations.
  - As mentioned above, the `f1 == f2` *lines* are already supported.

![n8_n16](https://user-images.githubusercontent.com/12679384/69210671-ddac0d00-0b9e-11ea-9966-f0d6ceba48d1.png)
![n8_n32](https://user-images.githubusercontent.com/12679384/69210679-e43a8480-0b9e-11ea-9e17-779246bdf4b1.png)
![n16_n32](https://user-images.githubusercontent.com/12679384/69210690-e997cf00-0b9e-11ea-9284-b6e7978ffbc2.png)

You can get the result of other cases with the following script:
```julia
using Gadfly, Colors

[...]

PR #166, opened 2020-01-14
Improve accuracy of `rem` with `Normed` types (e.g. `::Float32 % N0f32`)

### LINKED ISSUE #150: Rounding errors in `rem` of `Normed`

kimikage:
This fixes #150.

`Float32` and `Float64` versions can be unified, but the unified method may be difficult to read. So, I implemented them into two methods.
As I mentioned in https://github.com/JuliaMath/FixedPointNumbers.jl/issues/150#issuecomment-567035999, there are still numerical errors.

### DIFF
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -101,10 +101,24 @@ function _convert(::Type{N}, x::Tf) where {T, f, N <: Normed{T,f}, Tf <: Union{F
     return reinterpret(N, unsafe_trunc(T, yi >> (ex & bits)))
 end
 
-rem(x::T, ::Type{T}) where {T <: Normed} = x
-rem(x::Normed, ::Type{T}) where {T <: Normed} = reinterpret(T, _unsafe_trunc(rawtype(T), round((rawone(T)/rawone(x))*reinterpret(x))))
-rem(x::Real, ::Type{T}) where {T <: Normed} = reinterpret(T, _unsafe_trunc(rawtype(T), round(rawone(T)*x)))
-rem(x::Float16, ::Type{T}) where {T <: Normed} = rem(Float32(x), T)  # avoid overflow
+rem(x::N, ::Type{N}) where {N <: Normed} = x
+rem(x::Normed, ::Type{N}) where {T, N <: Normed{T}} = reinterpret(N, _unsafe_trunc(T, round((rawone(N)/rawone(x))*reinterpret(x))))
+rem(x::Real, ::Type{N}) where {T, N <: Normed{T}} = reinterpret(N, _unsafe_trunc(T, round(rawone(N)*x)))
+rem(x::Float16, ::Type{N}) where {N <: Normed} = rem(Float32(x), N)  # avoid overflow
+# Float32 and Float64 cannot exactly represent `rawone(N)` with `f` greater than
+# the number of their significand bits, resulting in rounding errors (issue #150).
+# So, we use another strategy for the large `f`s explained in:
+# https://github.com/JuliaMath/FixedPointNumbers.jl/pull/166#issuecomment-574135643
+function rem(x::Float32, ::Type{N}) where {f, N <: Normed{UInt32,f}}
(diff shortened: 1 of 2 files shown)


kimikage:
Subsequent commits: https://github.com/JuliaMath/FixedPointNumbers.jl/compare/master...kimikage:ctor_fixed

I want to modify some constructor-style conversion methods of `Normed` (e.g. issue #140), but that is not a bug fix.
So, what about releasing v0.8.0 after merging `ctor_fixed`?
In the following v0.9.0, let's change the arithmetic functions, i.e. let's start #152.

timholy:
I'm not certain I understand the strategy, so some comments would be useful. Does this essentially correspond to an affine rescaling? If so does it fix values near 1.0 at the expense of others? (The answers to these questions don't block merging, but let's document this a bit more.)

kimikage:
> Does this essentially correspond to an affine rescaling?

No. this is based on the "pure" linear transformation, so this is effective for almost all values other than the exception mentioned above.

First, the root of this problem is that while the width of `UInt32` is 32-bit, the mantissa of `Float32` is only 23+1-bit. This means that `rawone` of the type with `f > 24` cannot be exactly represented in `Float32`.
```julia
julia> rawone(Normed{UInt32, 24}) == Float32(0xFFFFFF)
true

julia> rawone(Normed{UInt32, 25}) == Float32(0x1FFFFFF)
false
```
Therefore, if `f <= 24`, we use the usual method.
```julia
    f <= 24 && return reinterpret(T, _unsafe_trunc(UInt32, round(rawone(T) * x)))
```
Well, the `rawone` is `2^f - 1`, i.e. 
```julia
x * rawone == x * 2^f  -  x
           == x * 2^(f + k - k)  -  x
           == x * 2^(f + k) * 2^-k  -  x * 2^(f + k) * 2^-(f + k))
```
where `k` is an arbitrary integer.
We calculate this with an integer type with the same bit width as `T` (i.e. 32-bit) in order to use registers efficiently. Since the bit width is finite, the choice of `k` has a trade-off between the accuracy and the overflow tolerance. I prioritized the accuracy and chose `f + k == 24`, i.e. the number of bits of mantissa.
Thus,
```julia
x * rawone == x * 2^24 * 2^(f - 24)  -  x * 2^24 * 2^-24
           == r * 2^(f - 24) - r * 2^-24
```
where `r == x * 2^24`. (`0x18 == 24`)
```julia
    r = _unsafe_trunc(UInt32, round(x * @f32(0x1p24)))
    reinterpret(T, r << UInt8(f - 24) - unsigned(signed(r) >> 0x18))
```

Do I make sense?

kimikage:
BTW, we may want to rename `T` to `N`. A month ago I seemed to follow the existing `rem`s.

timholy:
That's a great explanation. I'd say just insert a comment linking to that post and call it good.

Issue #156, opened 2019-12-26
[RFC] Acceptable input ranges


kimikage:
When fixing the issue #102 (PR #131), I changed the acceptable input range for `Normed` as follows:
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7ad0f0c460961024191f672332fcf3d7cef655d1/src/normed.jl#L69-L74

`(typemax(T)-rawone(U))/rawone(U)+1` essentially means `typemax(T)/rawone(U)`.

The decision is based, for example, on the idea that the range of `N0f8` should be [0, 1]. In particular, regarding the lower bound, I believe that the `Normed`s should not accept negative inputs since the current `Normed`s are unsigned types.

However, the following is also true:
```julia
julia> round(UInt8, 1.0019607843137253 * 255) # typemax(UInt8)/rawone(N0f8) == 1
0xff

julia> round(UInt8, -0.00196078431372549 * 255)
0x00
```
So, I don't know whether the acceptable input range of `Q0f7` should be [-1, 127/128] , or [-128.5/128, 127.5/128).
Accepting inputs ​​less than `typemin` or greater than `typemax` causes a bit confusing, and it is nonsense especially if the input type is not based on the binary numeral system. On the other hand, it seems to be a desirable property that the difference between the upper and lower bounds is a power of two.

kimikage:
BTW, I do not fully understand the original intent of this test.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7ad0f0c460961024191f672332fcf3d7cef655d1/test/fixed.jl#L95-L100

kimikage:
For the time being, I decided to allow the wider range than [`typemin`, `typemax`].

PR #169, opened 2020-01-16
Add support for `Normed` construction from `Rational` (Fixes #157)

### LINKED ISSUE #157: Problems with conversions from `Rational`

kimikage:
This also fixes the overflow problem with `Fixed` construction from `Rational`. (Fixes #157)

Note the input range that `Fixed` allows (cf. issue #156).

### DIFF
--- a/src/fixed.jl
+++ b/src/fixed.jl
@@ -62,7 +62,13 @@ function _convert(::Type{F}, x::AbstractFloat) where {T, f, F <: Fixed{T,f}}
 end
 
 function _convert(::Type{F}, x::Rational) where {T, f, F <: Fixed{T,f}}
-    F(x.num)/F(x.den) # TODO: optimization and input range checking
+    xmin = widemul(denominator(x), widen1(T)(typemin(T)) << 0x1 - 0x1)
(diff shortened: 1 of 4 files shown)


kimikage:
I chose the input range just because it is looser (i.e. it has a low impact of the change).
@timholy, if it is not too much trouble, please tell me what the input range should be.

kimikage:
Of course, anyone's opinion is fine. I'm worried about setting the specifications against the thoughts of "silent" users. However, I also think that this is not a good place for consensus building.

kimikage:
I cannot see the negative effect unless we release it. I'm concentrating my efforts on releasing v0.8.0.

PR #170, opened 2020-01-21
Modify Fixed constructors

kimikage:
This adds explicit input range checks. As a result, the constructors throw `ArgumentError` instead of `InexactError` for out-of-range inputs.

### DIFF
--- a/src/fixed.jl
+++ b/src/fixed.jl
@@ -48,17 +48,39 @@ fracmask(x::Fixed{T,f}) where {T, f} = ~intmask(x) # Signed
 
 # constructor-style conversions
 function _convert(::Type{F}, x::Fixed{T2,f2}) where {T, T2, f, f2, F <: Fixed{T,f}}
-    y = round(((1<<f)/(1<<f2))*reinterpret(x)) # FIXME: avoid overflow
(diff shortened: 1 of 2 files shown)


kimikage:
Thank you for your review.

I think there are roughly two types of user-defined float types. One is a thin wrapper like `struct MyFloat<:AbstractFloat; v::Float64; end` and the other is a more complicated type with special functionality.
For the former, the current code requires the implementation of `trunc` and `rem`, and promotion rules with integers. That results in unnecessary implementation costs and execution costs. Therefore, the end users or package developers should implement conversion code using its inner type.

I initially thought of using `floattype(F)` when `promote_type(floattype(F), typeof(x)) == floattype(F)`. Perhaps it can provide a faster-and-safe way than the `big` method. However, I think it is a "premature optimization". The `Float32`/`Float64` <--> `Integer` conversions are not slow but not fast, and above all, those conversion methods can throw exceptions.

If we actually find that it is better to use `floattype(F)`, then, we should use it.

kimikage:
The speedup can be done after releasing v0.8.0, so I will merge this first. With this PR, I am ready to release v0.8.0.

PR #177, opened 2020-03-17
Support `floattype(Rational)` and require <:AbstractFloat for fallback

timholy:
The more "dangerous" ~~(breaking)~~ part of this is that it seems to be a non-sequitur to allow `floattype` to fall back to returning a type that is not an `AbstractFloat`. ~~Consequently this will now return a `MethodError` under more circumstances.~~

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -156,6 +156,12 @@ floattype(::Type{X}) where {T <: Integer, X <: FixedPoint{T}} = Float64
 floattype(::Type{X}) where {T <: Integer, X <: Rational{T}} = typeof(zero(T)/oneunit(T))
 floattype(::Type{X}) where {T <: LongInts, X <: FixedPoint{T}} = BigFloat
 
+# Non-Real types
+floattype(::Type{Complex{T}}) where T = Complex{floattype(T)}
+floattype(::Type{<:Irrational}) = Float64
+floattype(::Type{Base.TwicePrecision{Float64}}) = Float64    # wider would be nice, but hardware support is paramount
+floattype(::Type{Base.TwicePrecision{T}}) where T<:Union{Float16,Float32} = widen(T)
+
 float(x::FixedPoint) = convert(floattype(x), x)
 
 function minmax(x::X, y::X) where {X <: FixedPoint}
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -106,34 +106,54 @@ floatmax(::Type{T}) where {T <: FixedPoint} = typemax(T)
 
 
 """
-    floattype(::Type{T})
+    floattype(::Type{T})::Type{<:AbstractFloat}
 
-Return the minimum float type that represents `T` without overflow to `Inf`.
+Return a minimal type suitable for performing computations with instances of type `T` without integer overflow.
 
-# Example
+The fallback definition of `floattype(T)` applies only to `T<:AbstractFloat`.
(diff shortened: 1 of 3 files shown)


timholy:
OK, this has been pretty extensively reworked. First, this issues a deprecation warning rather than an error. Second, it clarifies when it's OK to extend this to support non-AbstractFloats.

I'm contemplating releasing this as part of the 0.8 series, rather than as the beginning of a 0.9 series. I've traditionally viewed depwarns as more breaking than others have, but I'm beginning to be more cautious about tagging-churn in the dependencies. Thoughts?

timholy:
A demo in the OP of https://github.com/JuliaGraphics/ColorTypes.jl/pull/177 needed to add support for `floattype(::Type{<:Complex})`. I am not certain whether that would be worth adding here, but if so this would be a good PR for it.

kimikage:
It seems that what `floattype` for non-`Real` types should return are not well defined.
For example, I am very confused what the `floattype` of `TwicePrecision{T}` should be. `TwicePrecision{T}` is not a `Real`, but `Real`-like. Well, `Gray{T}` is also `Real`-like.

`Complex` is also complex. `Complex` is a scalar type in some contexts, and also is a vector type in other contexts.

kimikage:
`floattype(::Type{<:AbstractIrrational}) = Float64` Or, `BigFloat`? :dizzy_face:

kimikage:
> I'm contemplating releasing this as part of the 0.8 series, rather than as the beginning of a 0.9 series.

FixedPointNumbers is still in v0 series, so I think it is OK. However, v0.8 is starting to be used as the blocking by ColorTypes and Colors has been lifted. Therefore, it is better to make decisions early.

timholy:
I thought about adding it for `TwicePrecision` but couldn't come up with a test that leveraged "real" functionality of this package (that is, doing something other than calling `floattype` directly). But perhaps I should break the horror and make a decision. Most recent commit contains my answers to these questions

timholy:
It's long overdue, but I'm finally ready to submit the implementation of https://github.com/JuliaGraphics/ColorVectorSpace.jl/issues/126. I aim to merge dependencies in FixedPointNumbers and ColorTypes first, so expect a couple of PRs (like this one) that have languished to move into the "let's merge this" mode.

src/FixedPointNumbers.jl
```diff
@@ -106,36 +106,62 @@ floatmax(::Type{T}) where {T <: FixedPoint} = typemax(T)
...
 floattype(::Type{T}) where {T <: Integer} = Float64
 floattype(::Type{T}) where {T <: LongInts} = BigFloat
 floattype(::Type{X}) where {T <: ShortInts, X <: FixedPoint{T}} = Float32
 floattype(::Type{X}) where {T <: Integer, X <: FixedPoint{T}} = Float64
+floattype(::Type{X}) where {T <: Integer, X <: Rational{T}} = typeof(zero(T)/oneunit(T))
 floattype(::Type{X}) where {T <: LongInts, X <: FixedPoint{T}} = BigFloat
 
+# Non-Real types
+floattype(::Type{Complex{T}}) where T = Complex{floattype(T)}
+floattype(::Type{<:Irrational}) = Float64
```
kimikage, line 161:
```suggestion
floattype(::Type{<:AbstractIrrational}) = Float64
```
This is not clear, but in terms of precision, `Irrational` and other subtypes of `AbstractIrrational` should have the same result.

BTW, `AbstractIrrational` is a `Real` type. (see the caption comment "# Non-Real types")

PR #180, opened 2020-05-04
Reduce invalidations of other methods

timholy:
These changes, combined with more extensive changes to Julia itself, greatly reduce latency stemming from loading FixedPointNumbers. Ref https://github.com/JuliaLang/julia/pull/35733.

There will be very little benefit to this on its own, but we can at least find out if it works across Julia versions.

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -48,7 +48,8 @@ rawtype(::Type{X}) where {T, X <: FixedPoint{T}} = T
 *(x::Real, ::Type{X}) where {X <: FixedPoint} = _convert(X, x)
 
 # constructor-style conversions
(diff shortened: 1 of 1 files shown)


kimikage:
```julia
julia> sizeof(Fixed{BigInt,10})
8
```
:stuck_out_tongue_closed_eyes:
This is off-topic but it doesn't seem to be well documented what sizeof should return. (https://docs.julialang.org/en/v1.4/base/base/#Base.sizeof-Tuple{Type})

kimikage:
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/f0f9e74a37ca6170be615f0796d3105f1b22bbe6/src/FixedPointNumbers.jl#L6

We will not need `sizeof, ` anymore.

timholy:
Oof. That's not good. I'm surprised that `Core.sizeof` does that.

PR #183, opened 2020-06-18
Fix reductions in Statistics

timholy:
Formerly we were choosing an inconsistent eltype for reductions performed by Statistics. Since that's a stdlib that comes for "free," it seems we should support it on similar footing to `reduce` etc.

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -8,6 +8,8 @@ import Base: ==, <, <=, -, +, *, /, ~, isapprox,
              div, fld, rem, mod, mod1, fld1, min, max, minmax,
              rand, length
 
+import Statistics: _mean_promote
+
 using Base.Checked: checked_add, checked_sub, checked_div
 
 using Base: @pure
@@ -225,6 +227,7 @@ Base.mul_prod(x::FixedPoint, y::FixedPoint) = Treduce(x) * Treduce(y)
 Base.reduce_empty(::typeof(Base.mul_prod), ::Type{F}) where {F<:FixedPoint} = one(Treduce)
 Base.reduce_first(::typeof(Base.mul_prod), x::FixedPoint)  = Treduce(x)
 
+_mean_promote(x::Real, y::FixedPoint) = Treduce(y)
 
 """
     sd, ad = scaledual(s::Number, a)
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -8,7 +8,7 @@ import Base: ==, <, <=, -, +, *, /, ~, isapprox,
              div, fld, rem, mod, mod1, fld1, min, max, minmax,
              rand, length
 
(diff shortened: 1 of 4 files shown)


src/FixedPointNumbers.jl
```diff
@@ -8,6 +8,8 @@ import Base: ==, <, <=, -, +, *, /, ~, isapprox,
              div, fld, rem, mod, mod1, fld1, min, max, minmax,
              rand, length
 
+import Statistics: _mean_promote
```
kimikage, line 11: I don't think it is "free". Time is money.:stuck_out_tongue_closed_eyes:

https://github.com/JuliaLang/Statistics.jl/pull/25
The commit was merged on Apr. 14, 2020.

timholy: > Time is money

Development time or load time? Statistics is built into Julia itself, so I think it doesn't cost anything to import it here (any tiny amount of time is spent during precompilation). That's the sense of "free" I meant.

kimikage: I meant we have to wait a lot of time until `_mean_promote` is available in all the supported versions (i.e. until the release of v1.0.6 or the next LTS). Of course, this is a joke. :smile:

timholy: Very good. Nice catch, it's much better to make it conditional, see if this version is acceptable.

timholy:
Any objections or should I merge?

kimikage:
@timholy, are these #183, #180 and #177 in v0.8.1? Or in the v0.9 series?

As you mentioned, #177 is a somewhat dangerous change, but it shouldn't affect the "normal" functionality of downstream packages if there is no type piracy.

#186 is not a breaking change, and is not urgent. #187 is somewhat dangerous because it prohibits `copysign`/`flipsign` for `Normed`.

timholy:
#180 is in 0.8.1. None of the rest have been released yet. I think everything merged can be in the 0.8.x series, though, see https://github.com/JuliaMath/FixedPointNumbers.jl/pull/177#issuecomment-600555279.

kimikage:
Ah, I've already used v0.8.1, but I forgot it. :sweat_smile:
My interest is when to release v0.9. Originally in v0.9, I was planning to improve arithmetic.

timholy:
I'd say we could make a 0.8.2 release and then add any breaking changes.

Where do we stand on overflow? Without a clear consensus I'm reluctant to pick #143 up again. There's a part of me that thinks we should make all arithmetic immediately promote to `floattype` and just declare that these are storage-only numbers.

kimikage:
I will not drastically change the promotion rules unless there are concrete proposals. In other words, there has been no debate in the last six months, so I want to improve the arithmetic.

My first goal is to complete the set of `*_checked`, `wrapping_*`, and `saturating_*`.(cf. https://github.com/JuliaMath/FixedPointNumbers.jl/issues/152#issuecomment-647113657)

PR #186, opened 2020-06-22
Add `big()` and `rationalize()` for `FixedPoint`

### LINKED ISSUE #184: Adding `big(::FixedPoint)` and `rationalize(::FixedPoint)`
Conversion to `BigFloat` is already supported. However, `Base.big()` have not been overloaded.

In practical use, we should explicitly specify the `BigFloat`, but the short `big` may be useful for debugging.

Similarly, `rationalize()` have not been overloaded. Even though the name "rationalize" is longer than "Rational", I think supporting it is better for consistency.

kimikage:
Closes #184.

Perhaps, the users of `big` may need more accuracy than speed.(cf. https://github.com/JuliaMath/FixedPointNumbers.jl/issues/129#issuecomment-568863687)
I want to stop using the reciprocal technique below: (**Edit:** done.)
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7c90fb6b48867ff2fda1608b0b9c5b140e0ad963/src/normed.jl#L247
~For large arrays, using the reciprocal technique seems to allocate more memory and slow.~
In the first place, it seems that the constant folding is not applied. This seems to be essentially different from the following `Float64`'s case.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/7c90fb6b48867ff2fda1608b0b9c5b140e0ad963/src/FixedPointNumbers.jl#L99-L100

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -4,7 +4,7 @@ import Base: ==, <, <=, -, +, *, /, ~, isapprox,
              convert, promote_rule, show, isinteger, abs, decompose,
              isnan, isinf, isfinite,
              zero, oneunit, one, typemin, typemax, floatmin, floatmax, eps, reinterpret,
-             float, trunc, round, floor, ceil, bswap,
+             big, rationalize, float, trunc, round, floor, ceil, bswap,
              div, fld, rem, mod, mod1, fld1, min, max, minmax,
              rand, length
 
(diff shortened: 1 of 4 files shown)


kimikage:
```julia
julia> versioninfo()
Julia Version 1.4.2
Commit 44fa15b150* (2020-05-23 18:35 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, skylake)
```

Before:
```julia
julia> @benchmark BigFloat.(x) setup=(x=collect(rand(N0f8, 1000, 1000)))
BenchmarkTools.Trial:
  memory estimate:  328.06 MiB
  allocs estimate:  6000002
  --------------
  minimum time:     508.701 ms (38.52% GC)
  median time:      589.345 ms (37.99% GC)
  mean time:        631.865 ms (34.44% GC)
  maximum time:     829.537 ms (43.97% GC)
  --------------
  samples:          9
  evals/sample:     1

julia> BigFloat(5N32f32)
5.000000000000000000000000000000000000000000000000000000000000000000000000000069
```
After:
```julia
julia> @benchmark BigFloat.(x) setup=(x=collect(rand(N0f8, 1000, 1000)))
BenchmarkTools.Trial:
  memory estimate:  221.25 MiB
  allocs estimate:  4000002
  --------------
  minimum time:     325.089 ms (28.13% GC)
  median time:      450.361 ms (29.96% GC)
  mean time:        442.812 ms (31.00% GC)
  maximum time:     536.980 ms (40.39% GC)
  --------------
  samples:          12
  evals/sample:     1

julia> big(5N32f32)
5.0
```

kimikage:
I think there is room for improvement in the handling of `BigFloat`, but we should not need to improve it in this PR.

PR #190, opened 2020-07-07
Add checked, wrapping and saturating arithmetic for add/sub/neg

### LINKED ISSUE #152: Add `checked_add` and `checked_sub`

kimikage:
Closes #152.

The default arithmetic is still the wrapping arithmetic.

```julia
julia> using CheckedArithmetic

julia> @checked N0f8(0) + N0f8(1)
1.0N0f8

julia> @checked N0f8(0.5) + N0f8(0.5)
ERROR: OverflowError: 128 + 128 overflowed for type UInt8

julia> @checked 1N0f8 # Maybe some other time.
ERROR: MethodError: no method matching checked_mul(::Int64, ::Type{Normed{UInt8,8}})

julia> wrapping_add(0.5N0f8, 0.5N0f8)
0.0N0f8

julia> saturating_add(0.5N0f8, 0.5N0f8)
1.0N0f8
```
We may want to customize the overflow message.

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -12,7 +12,7 @@ import Base: ==, <, <=, -, +, *, /, ~, isapprox,
 import Statistics   # for _mean_promote
 import Random: Random, AbstractRNG, SamplerType, rand!
 
-using Base.Checked: checked_add, checked_sub, checked_div
+import Base.Checked: checked_neg, checked_add, checked_sub, checked_div
 
 using Base: @pure
 
@@ -35,7 +35,9 @@ export
 # "special" typealiases
     # Q and N typealiases are exported in separate source files
 # Functions
-    scaledual
+    scaledual,
+    wrapping_neg, wrapping_add, wrapping_sub,
+    saturating_neg, saturating_add, saturating_sub
 
 include("utilities.jl")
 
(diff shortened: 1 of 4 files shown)


kimikage:
Although I didn't intend to work on optimization in this PR, I added the specialized methods for unsigned saturation operation.
Even though SSE2 and AVX2 have 8-bit/16-bit saturation add/sub instructions, they are not used.

It may be better for Julia to support the LLVM's saturation arithmetic intrinsics. Perhaps that should have been discussed, but I couldn't find the latest info.

Anyway, `{add/sub}_with_overflow` are really terrible in vectorization. :sob: I don't like Matlab as a programming language, so I don't care the sighs of Matlab users. However I feel the need for a workaround.

kimikage:
I foresee that the 100% test of 16-bit types for division family will take time. (Of course, the division operation itself should not be slow.)

However, I feel like we can test the 8-bit types.

timholy:
That's kind of what I was thinking too.

kimikage:
I added the 8-bit `:thin` tests.
We may need to be serious about commonizing our tests (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/issues/139#issuecomment-664366361). :sweat_smile:

kimikage:
There are still opportunities to review the test code. I plan to improve the multiplication implementation in two or three PRs after this.

kimikage:
Checked arithmetic may throw an exception, so a slowdown is unavoidable. Instead of `checked` arithmetic, `wrapping` and `saturating` arithmetic can be used to provide some mitigation.

```julia
julia> x = rand(N0f8, 1000, 1000);

julia> y = typemax(N0f8) .- x;

julia> function checked_add_arrays(x, y)
           s = saturating_add.(x, y)
           all(wrapping_add.(x, y).==s) || throw(OverflowError("overflowed somewhere"))
           s
       end
checked_add_arrays (generic function with 1 method)

julia> @btime wrapping_add.($x, $y);
  86.299 μs (2 allocations: 976.70 KiB)

julia> @btime checked_add.($x, $y);
  898.900 μs (2 allocations: 976.70 KiB)

julia> @btime checked_add_arrays($x, $y);
  313.200 μs (6 allocations: 1.08 MiB)
```

PR #197, opened 2020-07-10
disable comment from codecov

johnnychen94:
Sometimes comment from codecov is annoying, and it can be disabled.

I'm not sure if this is wanted, just open it in case if you're not aware that this is doable.

Just merge or close it at will.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,1 @@
+comment: false


kimikage:
I'm concerned about the :x: due to superficial coverage regression, but not so annoyed with the comments. So, either is fine with me.

PR #194, opened 2020-07-08
Optimize `clamp`

kimikage:
This adds a specialized method for `clamp` with all arguments in the same `FixedPoint` type. This also supports the new `clamp(x, T)` method (cf. https://github.com/JuliaLang/julia/pull/34426).

This partially solves #179, but this is not helpful in cases where promotions occur.
cc: @johnnychen94

PR #187, opened 2020-06-28
Specialize sign-related functions

kimikage:
This is part of the issue #185.
**Edit:**
This prohibits `copysign`/`flipsign` for `Normed`. They should be confusing when the signed `Normed` is supported in the future. Furthermore, if signed `Normed` is available, we should not do unsafe sign operations for unsigned `Normed`.

Issue #185, opened 2020-06-22
Supplemental support for numerical and mathematical functions


kimikage:
In response to #179, I checked the support status and performance of some other functions. The (incomplete) list is shown below.

## To be supported
- [ ] `^(::FixedPoint, ::Integer)` for more accuracy
     - e.g. `N0f8(0.8)^3 != N0f8(64/125)`
- [ ] `clamp` (Issue #179)
- [ ] `bitstring`
- [ ] `big` (Issue #184)
- [ ] `copysign(x::Fixed, y)` and `flipsign(x::Fixed, y)` for speed performance
- [ ] `fma` and `muladd`
- [ ] `rationalize` (Issue #184)
- [ ] 3-arg `rem` and `div` (`cld`)
- [ ] `rem2pi` and `mod2pi`
- [ ] `signbit` for speed performance (for `Normed`)

## To be dropped
- [ ] `abs(::Normed)` specialization is not helpful. 
     - The fallback implementation works fine.
- [ ] `copysign(x::Normed, y)` and `flipsign(x::Normed, y)` should be throw error.

## TBD
- `maxintfloat` (`maxintfixed` and `minintfixed`?)
- `parse` and `tryparse` for fixed point numbers with suffix
- bit shift operators 
    - I think they are usuful but somewhat dangerous.
- `numerator` and `denominator`

kimikage:
Ah, I forgot `widemul`.

PR #198, opened 2020-07-10
Version 0.8.3

kimikage:
This PR is for displaying the progress. (cf. Issue #185)

- [x]  #186
- [x]  #187 (minor breaking change)
- [x]  #194 
- [x]  #195

I've confirmed that the invalidation problem in `Base` doesn't get too bad, but I haven't benchmarked with packages like `Plots`. (I'm not sure what the reasonable conditions are, as JuliaLang/julia has some PRs for invalidation problem, which haven't been merged yet.)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FixedPointNumbers"
 uuid = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
-version = "0.8.2"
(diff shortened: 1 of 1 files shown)


kimikage:
If nothing else needs to be fixed, I will release v0.8.3 and start introducing breaking changes as v0.9 material.

PR #203, opened 2020-07-19
Fix `_unsafe_trunc` for `BigFloat` on ARM

### LINKED ISSUE #202: MSB may be cleared in conversion from `BigFloat` on ARM
```julia
julia> BigFloat(typemax(UInt64))|>N63f1 # x86-64 / i686 
1.8446744073709552e19N63f1 

julia> BigFloat(typemax(UInt64))|>N63f1 # Aarch64 / ARMv7
9.223372036854776e18N63f1
```

_Originally posted by @kimikage in https://github.com/JuliaMath/FixedPointNumbers.jl/issues/200#issuecomment-660636301_

The following workaround is harmful to `BigFloat`.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/3f12729f76fa67682175c8dfe7d6c2ec1cd6c304/src/normed.jl#L330-L333

kimikage:
Fixes #202.

This also moves `_unsafe_trunc` to "src/utilities.jl". It has been used in ”fixed.jl" for 2 years.
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/39d46876dc641acf60c24fa4bb6f16575f3121d7/src/fixed.jl#L53

----

I didn't have the plans, but I'd like to release v0.8.4.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FixedPointNumbers"
 uuid = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
-version = "0.8.3"
+version = "0.8.4"
 
 [deps]
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
--- a/src/fixed.jl
+++ b/src/fixed.jl
@@ -54,7 +54,7 @@ end
 
 function _convert(::Type{F}, x::Integer) where {T, f, F <: Fixed{T,f}}
     if ((typemin(T) >> f) <= x) & (x <= (typemax(T) >> f))
-        reinterpret(F, unsafe_trunc(T, x) << f)
+        reinterpret(F, _unsafe_trunc(T, x) << f)
     else
         throw_converterror(F, x)
     end
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -323,12 +323,3 @@ end
     end
     :(Normed{$T,$f})
 end
-
-_unsafe_trunc(::Type{T}, x::Integer) where {T} = x % T
-_unsafe_trunc(::Type{T}, x) where {T}          = unsafe_trunc(T, x)
-if !signbit(signed(unsafe_trunc(UInt, -12.345)))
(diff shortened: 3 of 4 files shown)


kimikage:
The tests were passed on ARM.
https://travis-ci.com/github/kimikage/FixedPointNumbers.jl/builds/176247519

Since this package is low-level and Linux-aarch64 is in Tier 1, I think it might be a good idea to add aarch64 into the CI targets. However, the arm64 environment in Travis may differ from the practical environment because it does not support any extension instruction sets such as NEON.

johnnychen94:
> However, the arm64 environment in Travis may differ from the practical environment because it does not support any extension instruction sets such as NEON.

Even if it is not a perfect setting, adding `aarch64` as one CI target could definitely improve the code coverage and help catches potential issues.

kimikage:
My little concern is that the arm64 environment in Travis is a bit slow.

Maybe:
- GitHub Actions
  - { Linux, Windows, macOS } × { v1.0, v1, nightly } × { x86-64 } ∪ { (Windows, v1, x86) }
- Travis CI
  - { Linux } × { v1.0, v1, nightly } × { aarch64 }
- AppVeyor
  - Thank you for everything!

johnnychen94:
Windows x86 is no-longer at tier 1 support, so it makes sense to not include it in CI. I'm not 100% sure of it, though.

https://github.com/JuliaLang/www.julialang.org/pull/667

> { Linux, Windows, macOS } × { v1.0, v1, nightly } × { x86-64 }

I assume this is `arch = [x86, x64]`?

kimikage:
> I assume this is `arch = [x86, x64]`?

I meant `arch = x64` only.  I think testing on 32-bit x86 is necessary, but I don't think it's necessary to consider OS differences. Therefore, the OS for 32-bit x86 does not have to be Windows.

timholy:
I don't follow architectures carefully but I'd support adding a new one to tests. How long are we talking to run the tests on `aarch64`?

PR #204, opened 2020-07-21
[CI] use Github Actions and update test targets

johnnychen94:
This follows the discussion in #203 

* move CI workloads to Github Actions
* remove windows 32bit and say goodbye to appveyor
* test aarch64 for Julia 1.0, 1, and nightly in travis
* add linux 32bit for Julia 1


.github/workflows/UnitTest.yml
```diff
@@ -0,0 +1,55 @@
...
+      - master
+  pull_request:
+  schedule:
+    - cron: '20 00 1 * *'
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
johnnychen94, line 17: The default value for this is `true`.

If I get it correctly, Github Action currently doesn't support `allow_failure`, it would be nice to disable this to keep all jobs running.

timholy:
Both webhooks removed. Thanks for tackling this!

kimikage:
To make sure #201 and #203 work, I'll merge this first.
Generally, we should write `test_broken`s, but I will omit them. Regarding the failures on ARM, I think we can tolerate them depending on their importance, as long as the compilation succeeds. On the other hand, `allow_failure` should not be set to prevent oversight of errors.

kimikage:
@timholy, we are still using travis-ci.*org*. Why not move to travis-ci.*com*?

johnnychen94:
For public projects, there seems no difference. We could do the switch when there's real trouble using it.

kimikage:
I didn't expect [the pricing model change](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) before the [the migration deadline](https://docs.travis-ci.com/user/migrate/open-source-repository-migration#frequently-asked-questions). :sweat_smile:

I can't predict how much of a hassle it will be to re-request credits, but should we migrate to .com first?

PR #206, opened 2020-07-24
Use sampler-based Random API

### LINKED ISSUE #125: poor performance

### LINKED ISSUE #196: [RFC] Use of Random API
Currently, the random number array of the `FixedPoint` is generated by `reinterpret`ing the random number array of its rawtype.
The generation is very fast, but the use of `ReinterpretArray` brings poor performance. (cf. #125)

Julia v1.0 and later, the sampler-based [Random API](https://docs.julialang.org/en/v1/stdlib/Random/#Generating-random-values-of-custom-types-1) is provided. By using the Random API, we can generate random numbers more flexibly. The disadvantage is that the generation of random arrays is much slower.

Current:
```julia
julia> @btime rand(N0f8, 1000, 1000); # ReinterpretArray
  139.100 μs (3 allocations: 976.73 KiB)

julia> @btime collect(rand(N0f8, 1000, 1000));
  2.479 ms (5 allocations: 1.91 MiB)

julia> @btime replace!(_->rand(N0f8), Array{N0f8}(undef, 1000, 1000));
  2.744 ms (2 allocations: 976.70 KiB)
```
Using the newer Random API:
```julia
julia> @btime rand(rng, N0f8, 1000, 1000) setup=(rng=Random.MersenneTwister(1234)); # we can specify the RNG
  1.231 ms (2 allocations: 976.70 KiB)
```

Do you have any thoughts?

kimikage:
Closes #196, Fixes #125

This avoids the performance problem with `ReinterpretArray`. ~On the other hand, the time for generating the random number array with the default RNG is significantly increased.~

~I chose a minimalistic and straightforward implementation for now.~

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
 version = "0.8.4"
 
 [deps]
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -7,9 +7,10 @@ import Base: ==, <, <=, -, +, *, /, ~, isapprox,
              big, rationalize, float, trunc, round, floor, ceil, bswap, clamp,
              div, fld, rem, mod, mod1, fld1, min, max, minmax,
              signed, unsigned, copysign, flipsign, signbit,
-             rand, length
+             length
 
 import Statistics   # for _mean_promote
+import Random: Random, AbstractRNG, SamplerType, rand!
 
 using Base.Checked: checked_add, checked_sub, checked_div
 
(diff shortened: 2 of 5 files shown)


timholy:
How bad is the regression for generation? `unsafe_wrap` isn't that scary, personally I think we should use it.

kimikage:
I added a method using `unsafe_wrap`. I don't think `GC.@preserve` is needed there, but I'm not sure.

kimikage:
Thank you for your prompt review.

Issue #78, opened 2017-02-23
Define `::Integer * ::FixedPoint`?


TotalVerb:
```julia
julia> 2 * 0.5Q1f14
ERROR: InexactError()
Stacktrace:
 [1] convert at /home/fengyang/.julia/v0.6/FixedPointNumbers/src/fixed.jl:42 [inlined]
 [2] promote at ./promotion.jl:174 [inlined]
 [3] *(::Int64, ::FixedPointNumbers.Fixed{Int16,14}) at ./promotion.jl:247
```

This is happening because `2` is not representable in Q1f14, even though the result `1` would be. I think it is avoidable by specializing `::Integer * ::FixedPoint` (and `::FixedPoint * ::Integer`?) directly, but I don't have much experience with this. If this is the right thing to do, I'll make a PR.

vchuravy:
Yes this is due to our promote rules... I think implementing `::Integer * ::FixedPoint` is fine, but we need to think about whether we want to check for overflow.

kimikage:
I misunderstood this issue. This is a problem with promotion rules only for `Fixed`, not for `Normed`. Since the birth of `Normed` (formerly `UFixed`), i.e. PR #2, it has been promoted to `Float32`/`Float64`.

I thought the `checked_mul` could avoid unnecessary errors, but that would lead to further inconsistencies. :confused:

PR #207, opened 2020-07-24
Commonize promotion rules

### LINKED ISSUE #192: [RFC] Changing the promotion type of `Integer` and `Fixed` to a float type.
I was considering introducing `checked_mul` for Issue #78. However, I noticed that there are differences in promotion rules between `Fixed` and `Normed`.

```julia
julia> 2 * 0.5N2f14 # promotion to Float32
1.000061f0

julia> 2 * 0.5Q1f14 # promotion to Q1f14 (issue #78)
ERROR: ArgumentError: Fixed{Int16,14} is a 16-bit type representing 65536 values from -2.0 to 1.99994; cannot represent 2
```

Perhaps this is due to the current (historical) implementation where `Fixed` is signed and `Normed` is unsigned, but IMO, it seems practical to promote them to float types, and not to add `::Integer * ::FixedPoint`.

kimikage:
Closes #192, #126, #78

This makes the promotion rules for `Fixed` consistent with the rules used within `Normed`:
- `Fixed` and `Integer` are promoted to `floattype`
- `Fixed` and `AbstractFloat` are promoted to `floattype` with sufficient precision

This adds the rule:
- `Fixed` and `Normed` are promoted to `floattype`

This also stops using `@generated` function.

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -311,6 +311,35 @@ include("normed.jl")
 include("deprecations.jl")
 const UF = (N0f8, N6f10, N4f12, N2f14, N0f16)
 
+# Promotions
+promote_rule(::Type{X}, ::Type{Tf}) where {X <: FixedPoint, Tf <: AbstractFloat} =
+    promote_type(floattype(X), Tf)
+
+# Note that `Tr` does not always have enough domains. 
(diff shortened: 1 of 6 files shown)


kimikage:
I ran `rebase master`.

I'm not sure how `Fixed` is used in the wild. All I can do now is pray that this breaking change does not cause a fatal problem.

PR #208, opened 2020-08-04
Systemize fixed point types to be tested

kimikage:
cf. https://github.com/JuliaMath/FixedPointNumbers.jl/issues/139#issuecomment-664006515

This does not commonize the tests between `Fixed` and `Normed`, but does prepare for the commonization.

### DIFF
--- a/test/common.jl
+++ b/test/common.jl
@@ -1,0 +1,130 @@
+using FixedPointNumbers, Statistics, Test
+using FixedPointNumbers: bitwidth, rawtype, nbitsfrac
+
+"""
+    target(X::Type, Ss...; ex = :default)
+
+Return a generator which enumerates the target types for testing.
+
+# Arguments
+- `X`: target base type
+- `Ss`: symbols for specifying the target raw types
+  - `:i*` : a `Signed` type if `X === Fixed`, or an `Unsigned` type if `X === Normed`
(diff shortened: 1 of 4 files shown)


timholy:
Due to code movement this won't be easy to review, and it doesn't look like it's much in need anyway. So just go ahead and merge when you're ready.

kimikage:
Thank you for your response. Most of the code movement is not important, but I don't want to mess with the arrangement every time I add a test set. :sweat_smile:

The essential change in this PR is https://github.com/JuliaMath/FixedPointNumbers.jl/pull/208/commits/2a4a61ba207d6f75e83b231e41a22e789364a3d9.
TBH, I don't really like the new `target` function because I think the test code should be simple. However, test case "justification" is important, and this change saves more than 15 seconds of testing time on my local machine.

timholy:
Test performance is definitely important. Maybe add a little bit of commentary about the heavy/default/thin/light system? Just from reading it's not entirely obvious since the `target_f_series` is elsewhere.

kimikage:
I moved the `target_f_series` to "test/common.jl".  (it is contrary to the philosophy of  specialization, though. :smile:)
Also, I added the comment about the `heavy`/`default`/`light`/`thin` system as the docstring for `target_f`.

timholy:
LGTM!

PR #214, opened 2020-08-15
Improve overflow error message

kimikage:
This displays the fixed point type of the input in the message, not the rawtype. (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/pull/190#issue-445137281)
```julia
julia> checked_add(0.5N0f8, 0.5N0f8) # this PR
ERROR: OverflowError: 0.502N0f8 + 0.502N0f8 overflowed for type N0f8
```
Of course, more information could be added to the error message, but I think the error message in the constructor is sufficient to supplement it.
```julia
julia> 1.004N0f8
ERROR: ArgumentError: N0f8 is an 8-bit type representing 256 values from 0.0 to 1.0; cannot represent 1.004
```

This also avoids the stall with `checked_{add/sub}` on Windows. (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/pull/190#issuecomment-674334221) TBH, this is the real purpose of this PR. :sweat_smile:
```julia
julia> @btime checked_add.($x, $y);
  898.900 μs (2 allocations: 976.70 KiB) # before (on Windows)
  577.699 μs (2 allocations: 976.70 KiB) # after (on Windows)
  678.800 μs (2 allocations: 976.70 KiB) # before (on Linux)
  679.500 μs (2 allocations: 976.70 KiB) # after (on Linux)
```

### DIFF
--- a/test/fixed.jl
+++ b/test/fixed.jl
@@ -275,8 +275,8 @@ end
         xs = typemin(F):eps(F):typemax(F)
         fneg(x) = -float(x)
         @test all(x -> wrapping_neg(wrapping_neg(x)) === x, xs)
-        @test all(x -> saturating_neg(x) == clamp(fneg(x), F), xs)
-        @test all(x -> !(typemin(F) < fneg(x) < typemax(F)) ||
+        @test all(x -> saturating_neg(x) === clamp(fneg(x), F), xs)
+        @test all(x -> !(typemin(F) <= fneg(x) <= typemax(F)) ||
                        wrapping_neg(x) === checked_neg(x) === fneg(x) % F, xs)
     end
 end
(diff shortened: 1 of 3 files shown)


test/fixed.jl
```diff
@@ -275,8 +275,8 @@ end
         xs = typemin(F):eps(F):typemax(F)
         fneg(x) = -float(x)
         @test all(x -> wrapping_neg(wrapping_neg(x)) === x, xs)
-        @test all(x -> saturating_neg(x) == clamp(fneg(x), F), xs)
-        @test all(x -> !(typemin(F) < fneg(x) < typemax(F)) ||
+        @test all(x -> saturating_neg(x) === clamp(fneg(x), F), xs)
+        @test all(x -> !(typemin(F) <= fneg(x) <= typemax(F)) ||
```
kimikage, line 279: I had excluded `typemin` and `typemax` with the intention of writing generic code for 100% testing. They have a risk of causing inconsistencies due to the rounding errors in floating point operations.
However, I decided against supporting non-8-bit types in 100% tests for multiplication. So we can safely include `typemin` and `typemax` for 8-bit types.

Issue #219, opened 2020-08-20
Problems with `rem` for `Fixed`


kimikage:
As mentioned in https://github.com/JuliaMath/FixedPointNumbers.jl/issues/211#issuecomment-673995804, the `rem` for `Fixed` has problems. 
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/735b7f919310d604ed49471aa6a645a72eb6caf5/src/fixed.jl#L109
I plan to specialize the multiplication for `Fixed` after the merging PR #213. As part of the specialization, I will improve the `rem` for `Fixed`, but since it is not the main topic of the specialization, I opened a separate issue for `rem`.

## Speed performance
Simply put, `rem` for `Fixed` is too slow.
```julia
julia> x_f32 = rand(Float32, 1000, 1000);

julia> @btime $x_f32 .% Q0f7;
  8.012 ms (2 allocations: 976.70 KiB)

julia> @btime $x_f32 .% N0f8;
  123.300 μs (2 allocations: 976.70 KiB)
```

## Throwing errors
Though this is one of the causes of the slowdown, the `rem` can throw an error.
```julia
julia> 2.0^63 % Q0f7 # Due to `Integer()`
ERROR: InexactError: Int64(9.223372036854776e18)
```
The `rem` for `Normed` prioritizes the speed and allows for inaccurate values to be returned in the overflow.

## Unnecessary type conversion
We need the specialized method.
```julia
julia> typemax(Q0f63) % Q0f63
ERROR: ArgumentError: Fixed{Int64,63} is a 64-bit type representing 0 values from -1.0 to 1.0; cannot represent 1
```

kimikage:
FYI, the behavior of `unsafe_trunc` on x86 CPUs is described below.

> If the value is not representable by T, an arbitrary value will be returned.

https://docs.julialang.org/en/v1/base/math/#Base.unsafe_trunc

On x86 CPUs, a family of `(v)cvttsd2si` instructions is used to convert floating point numbers to integers. The integer types are signed, and `Int32` and `Int64` are supported, i.e. the conversion to 8-/16-bit integers is done via Int32. Also, `0x80000000`/`0x80000000_00000000` is returned in case of overflow.
[...]

PR #213, opened 2020-08-14
Optimize multiplication for Normed

kimikage:
This adds `wrapping_mul`, `saturating_mul` and `checked_mul` binary operations. However, this does not specialize them for `Fixed` and does not change [`*` for `Fixed`](https://github.com/JuliaMath/FixedPointNumbers.jl/blob/5dcaf979ac148012d7d95b018232c9a86d7416a9/src/fixed.jl#L103).

This applies the wrapping arithmetic as the default arithmetic of `*` for the abstract `FixedPoint`, but as mentioned above, the `*`for `Fixed` is not affected by the default arithmetic. Furthermore, `*` for `Normed` overrides the default arithmetic with the "checked" arithmetic for backward compatibility. (Strictly speaking, the error type is changed from `ArgumentError` to `OverflowError`.)

This replaces most of Normed's implementation of multiplication with integer operations. This improves the speed in many cases and the accuracy in some cases.

Fixes #174

PR #220, opened 2020-08-21
Specialize multiplication for `Fixed`

### LINKED ISSUE #173: Changing rounding mode of "mul" (`*`) for `Fixed` to `RoundNearest`
The current rounding mode of "mul"(`*`) for `Fixed` is `RoundNearestTiesUp`. (The comment "rounding up" is ambiguous, though.) In addition, the implementation with `RoundDown` is commented out. (The comment "truncation" is also ambiguous, though.)
https://github.com/JuliaMath/FixedPointNumbers.jl/blob/cb29336982309076894d63a0748f9ead13c672bc/src/fixed.jl#L76-L79

However, considering the future addition of `checked_mul`, I think it would be better to change the rounding mode to the default (i.e. `RoundNearest`) for consistency.

On the other hand, I would also like to have the interfaces to implementations with other rounding modes, for speed. The problem is that there is no official interface that takes a rounding mode as an argument. The third argument of `Base.*()` collides with `z` in `x * y * z` (Of course, there are no arithmetic operations for `RoundingMode`, though).

kimikage:
This is a sequel to PR #213. This specializes most of the multiplication for `Fixed` and avoids floating point operations.

A major change is that the rounding mode is changed from `RoundNearestTiesUp` to `RoundNearest`. The existing `RoundNearestTiesUp` and `RoundDown` (the latter was noted in a comment) modes are now supported by the new unexported function `mul_with_rounding`.

Unlike multiplication for `Normed`, the wrapping arithmetic is the default for `Fixed`.

This PR also improves `rem`.

Fixes #173, #219

### DIFF
--- a/src/fixed.jl
+++ b/src/fixed.jl
@@ -158,10 +158,18 @@ function checked_mul(x::F, y::F) where {T <: Union{Int8,Int16,Int32,Int64}, f, F
     F(div_2f(z, Val(Int(f))) % T, 0)
 end
 
-# with truncation:
-#*(x::Fixed{T,f}, y::Fixed{T,f}) = Fixed{T,f}(Base.widemul(x.i,y.i)>>f,0)
-# with rounding up:
-#*(x::Fixed{T,f}, y::Fixed{T,f}) where {T,f} = Fixed{T,f}((Base.widemul(x.i,y.i) + (one(widen(T)) << (f-1)))>>f,0)
+function mul_with_rounding(x::F, y::F, ::RoundingMode{:Nearest}) where {F <: Fixed}
+    wrapping_mul(x, y)
+end
+function mul_with_rounding(x::F, y::F, ::RoundingMode{:NearestTiesUp}) where
+                          {T <: Union{Int8,Int16,Int32,Int64}, f, F <: Fixed{T, f}}
+    z = widemul(x.i, y.i)
+    F(((z + (oftype(z, 1) << f >>> 1)) >> f) % T, 0)
+end
+function mul_with_rounding(x::F, y::F, ::RoundingMode{:Down}) where
+                          {T <: Union{Int8,Int16,Int32,Int64}, f, F <: Fixed{T, f}}
+    F((widemul(x.i, y.i) >> f) % T, 0)
+end
 
 /(x::Fixed{T,f}, y::Fixed{T,f}) where {T,f} = Fixed{T,f}(div(convert(widen(T), x.i) << f, y.i), 0)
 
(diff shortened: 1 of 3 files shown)


kimikage:
## Benchmark
See [`Normed` version](https://github.com/JuliaMath/FixedPointNumbers.jl/pull/213#issuecomment-674039502)

```julia
x_q0f7  = collect(rand(Q0f7,  1000, 1000)); y_q0f7  = collect(rand(Q0f7,  1000, 1000));
x_q0f15 = collect(rand(Q0f15, 1000, 1000)); y_q0f15 = collect(rand(Q0f15, 1000, 1000));
x_q3f4  = collect(rand(Q3f4,  1000, 1000)); y_q3f4  = collect(rand(Q3f4,  1000, 1000));
x_q3f12 = collect(rand(Q3f12, 1000, 1000)); y_q3f12 = collect(rand(Q3f12, 1000, 1000));
z_q3f4  = Q3f4.( rand(1000, 1000));
z_q3f12 = Q3f12.(rand(1000, 1000));


julia> @btime $x_q0f7 .* $y_q0f7;
  100.800 μs (2 allocations: 976.70 KiB)

julia> @btime $x_q0f15 .* $y_q0f15;
  806.899 μs (2 allocations: 1.91 MiB)

julia> @btime saturating_mul.($x_q3f4, $y_q3f4);
  162.899 μs (2 allocations: 976.70 KiB)

julia> @btime saturating_mul.($x_q3f12, $y_q3f12);
  928.600 μs (2 allocations: 1.91 MiB)

julia> @btime checked_mul.($x_q3f4, $z_q3f4);
  1.353 ms (2 allocations: 976.70 KiB)

julia> @btime checked_mul.($x_q3f12, $z_q3f12);
  1.666 ms (2 allocations: 1.91 MiB)
```

cf. v0.8.4 (`RoundNearestTiesUp` version)
```julia
julia> @btime $x_q0f7 .* $y_q0f7;
  77.900 μs (2 allocations: 976.70 KiB)

julia> @btime $x_q0f15 .* $y_q0f15;
  758.600 μs (2 allocations: 1.91 MiB)
```

I feel we can improve it a bit more, but I don't plan to work on optimizing it as a priority.

kimikage:
As an additional note for later readers, the default `*`, i.e. `wrapping_mul`, is slightly slower due to the change in rounding mode. Instead, at least for `Fixed{Int8}` and `Fixed{Int16}`, the results will agree with the float.

As for the `rem`, the speed is greatly improved by allowing incorrect results from overflow as well as `Normed`.

PR #223, opened 2020-08-26
Fix `% Normed{UInt32}` on ARM and Improve `NaN % FixedPoint`

### LINKED ISSUE #134: `rem` returns zero for negative float input on ARMv7

kimikage:
Fixes #134 (https://github.com/JuliaMath/FixedPointNumbers.jl/issues/134#issuecomment-680622777)
```julia
julia> 1.0 % N0f32
0.4999999999N0f32 # before (on ARM)
1.0N0f32 # after
```
This also adds tests for `NaN`/`Inf`. This clarifies the behavior regarding `NaN` and `Inf`, which had not been explicitly defined.

This reduces the environment-/optimization-dependent instability of the results for `NaN`. 
```julia
  Expression: NaN % N === NaN32 % N === zero(N)
   Evaluated: 65537.0N16f16 === 65537.0N16f16 === 0.0N16f16 # I could not reproduce this on my local machine.
```
(cf. https://github.com/JuliaMath/FixedPointNumbers.jl/runs/1029971789?check_suite_focus=true#step:5:26)
However, the `rem` is still affected by the undefined behavior of `unsafe_trunc`.

### DIFF
--- a/src/fixed.jl
+++ b/src/fixed.jl
@@ -103,10 +103,18 @@ function rem(x::Fixed, ::Type{F}) where {T, f, F <: Fixed{T,f}}
 end
 rem(x::Integer, ::Type{F}) where {T, f, F <: Fixed{T,f}} = F(_unsafe_trunc(T, x) << f, 0)
 function rem(x::Real, ::Type{F}) where {T, f, F <: Fixed{T,f}}
-    y = _unsafe_trunc(promote_type(Int64, T), round(x * @exp2(f)))
+    if bitwidth(T) < 32
+        Ti = T
+    else
+        isfinite(x) || return zero(F)
+        Ti = promote_type(Int64, T)
+    end
+    Tf = floattype(F)
+    y = _unsafe_trunc(Ti, round(x * Tf(@exp2(f))))
     reinterpret(F, _unsafe_trunc(T, y))
 end
 function rem(x::BigFloat, ::Type{F}) where {T, f, F <: Fixed{T,f}}
+    isfinite(x) || return zero(F)
     reinterpret(F, _unsafe_trunc(T, round(x * @exp2(f))))
 end
 
--- a/src/normed.jl
+++ b/src/normed.jl
@@ -110,18 +110,23 @@ end
 
 rem(x::N, ::Type{N}) where {N <: Normed} = x
 rem(x::Normed, ::Type{N}) where {T, N <: Normed{T}} = reinterpret(N, _unsafe_trunc(T, round((rawone(N)/rawone(x))*reinterpret(x))))
-rem(x::Real, ::Type{N}) where {T, N <: Normed{T}} = reinterpret(N, _unsafe_trunc(T, round(rawone(N)*x)))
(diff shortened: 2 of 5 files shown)


kimikage:
This is not the main topic of this PR, but the speed performance of `% Fixed{Int8}` was improved.
```julia
julia> x_f32 = rand(Float32, 1000, 1000);

julia> @btime $x_f32 .% Q0f7;
  8.012 ms (2 allocations: 976.70 KiB)   # v0.8.4
  762.600 μs (2 allocations: 976.70 KiB) # PR #220 
  125.699 μs (2 allocations: 976.70 KiB) # this PR
```
Of course this makes it easier to overflow as a trade-off, but still has enough range for practical purposes.

kimikage:
This change is somewhat ugly, but I will merge this first as it is needed to implement the "new" division.

PR #225, opened 2020-09-03
Fix ambiguity test on Julia v1.6.0-DEV

kimikage:
cf. JuliaLang/julia PR #36962

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,6 +1,10 @@
 using FixedPointNumbers, Test
 
-@test isempty(detect_ambiguities(FixedPointNumbers, Base, Core))
+if VERSION >= v"1.6.0-DEV.816" # JuliaLang/julia #36962
+    @test isempty(detect_ambiguities(FixedPointNumbers))
+else
+    @test isempty(detect_ambiguities(FixedPointNumbers, Base, Core))
+end
 
 @testset "normed" begin
     include("normed.jl")


johnnychen94:
FYI https://github.com/JuliaTesting/Aqua.jl can be used to automate this kind of meta quality test

kimikage:
Yes. However, keeping up with the changes in the nightly build is generally a necessary task, and what Aqua can automate is quite limited.

FixedPointNumbers is a "solid" package at the root of the JuliaImages ecosystem, so I can't expect any quality improvement by Aqua (in fact, this package should pass its tests). 

That said, from an assurance and traceability perspective, Aqua is helpful and I'm not opposed to its introduction.
Would you like to add the Aqua badge, or not?

johnnychen94:
The major advantage of Aqua shows up when you have tens of packages under watch and when all of them are hitting similar issues like this.

In any case, we don't need to immediately use Aqua; it is more of a personal taste and depends on how much maintenance work you have.

kimikage:
Further changes to the ambiguity computing may be made in the future due to latency regression and type intersection issues.

I would like to hold off on the decision to introduce Aqua until FixedPointNumbers v0.9.0 is ready for release. I have at least a couple of PRs to submit before the v0.9.0 release.

kimikage:
Ah, it's been reverted.:sweat_smile: (julia#37484)

PR #229, opened 2020-09-14
add doctest in unittest

johnnychen94:


timholy (approved):

kimikage:
I think this is a good change, but if we use Documenter, it might be a good idea to create a new document in the future. I have some performance tips for package developers.

PR #230, opened 2020-09-15
Add checked, wrapping and saturating arithmetic for `rem`/`mod`

### LINKED ISSUE #221: [RFC] Supporting checked arithmetic for division-related operations
While there is room for some improvement and discussion, I have prepared the implementations for checked arithmetic on addition and multiplication (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/issues/41#issuecomment-674682747).

However, the circumstances surrounding division-related operations are more complicated.

## Checked arithmetic for `/`

Since the result of the `/` with integer arguments is a floating point number, the checked version of `/` is not defined in `Base.Checked`. Unlike `*` and `mul`, `/` and `div` are different operations. So, `checked_div` is for `div`(`÷`) and not for `/`.
Shall we define `{checked/saturating/wrapping}_fdiv`? ("f" stands for "fractional", not "float" :smile:).

## Default arithmetic

Unlike addition and subtraction, the operations involved in integer division are basically "checked" arithmetic. It is not clear whether we should follow the integer style or the floating point style (which does not throw an exception). However, I personally think the checked arithmetic is a better choice for the default, since `FixedPoint` numbers cannot represent `Inf` or `NaN`.

## Division by zero in wrapping arithmetic

There are several options for the result of dividing by zero. Considering the distinction from saturating arithmetic, I think it is a good idea to return zero.

kimikage:
The default arithmetic for `rem` is still checked arithmetic.

`saturating_{rem|mod}` behaves somewhat strangely in the case of the overflow. (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/issues/221#issuecomment-689306004)
```julia
julia> x = typemin(Q0f7); y = -eps(Q0f7);

julia> saturating_div(x, y) # 128 --> 127
127

julia> saturating_rem(x, y) # != 0
-0.008Q0f7

julia> x == saturating_div(x, y) * y + saturating_rem(x, y)
true
```

This also adds the support for 3-arg `rem`. Only `RoundToZero`(default), `RoundDown`(`mod`), and `RoundUp` are supported. In `RoundUp` mode, `rem` causes `OverflowError` with unsigned (i.e. `Normed`) values. Otherwise, `rem`/`mod` does not cause an overflow.

This also adds dummy checked, wrapping and saturating `rem` for type modulus to prevent `MethodError` due to lexical replacement of `x % X`.

This completes the implementation of the checked arithmetic. Performance tweaks without changing behavior and the rest of #185 can be carried out after the release of v0.9.0.

Closes #221

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -13,7 +13,7 @@ import Statistics   # for _mean_promote
 import Random: Random, AbstractRNG, SamplerType, rand!
 
 import Base.Checked: checked_neg, checked_abs, checked_add, checked_sub, checked_mul,
-                     checked_div, checked_fld, checked_cld
+                     checked_div, checked_fld, checked_cld, checked_rem, checked_mod
 
 using Base: @pure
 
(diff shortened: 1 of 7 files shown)


timholy (approved):
I really like how much you've been able centralize here, nice work as always!

kimikage:
There is a bit more discussion about what to do with these arithmetic methods in v0.9.0, but I'll merge this first.
Although my plan is https://github.com/JuliaMath/CheckedArithmetic.jl/issues/9, @timholy may have a slightly different idea (cf. https://discourse.julialang.org/t/rfc-what-should-the-arithmetic-within-the-fixedpointnumbers-be/46697/36).

PR #235, opened 2020-10-02
Add support for `digits` keyword argument of `round()`

### LINKED ISSUE #234: Follow contract of floor/ceil
`floor` and `ceil` are supposed to return the same type as the input type:

```
floor(x) returns the nearest integral value of the same type as x that is less than or equal to x.
```

However, `floor(N0f8(0.33), digits=1)` returns `0.3f0` and `ceil(N0f8(0.33), digits=1)` returns `0.4f0` i.e. they return Float32.

I haven't tested other types, but since all calls to floor/ceil fallback to the Base definition of `floor(x; kwargs...) = round(x, RoundDown; kwargs...)` (RoundUp for ceil) and their callchains call `float(x)`, I suspect this is the case for all `Normed` types.

Found [here](https://discourse.julialang.org/t/how-to-floor-a-normed-fixed-point-number-in-an-idempotent-way/47510/5).

kimikage:
This only supports the non-negative `digits` option. All other options throw an error.
This provides specialized implementations only for `N0f8` and `Fixed{Int8}`. The other types use fallback with floating-point operations and therefore it is slower and less accurate.

Fixes #234

### DIFF
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -390,6 +390,35 @@ end
 
 signbit(x::X) where {X <: FixedPoint} = signbit(x.i)
 
+~(x::X) where {X <: FixedPoint} = X(~x.i, 0)
+
+function round(x::FixedPoint, r::RoundingMode=RoundNearest;
+               digits::Union{Nothing, Integer}=nothing,
+               sigdigits::Union{Nothing, Integer}=nothing,
+               base::Union{Nothing, Integer}=nothing)
+
+    d = digits === nothing ? 0 : Int(digits)
+
+    if base !== nothing && base != 10
+        throw(ArgumentError("`base` numbers other than 10 are not supported."))
+    elseif sigdigits !== nothing
+        throw(ArgumentError("`sigdigits` is not supported."))
+    elseif d < 0
+        throw(ArgumentError("negative `digits` is not supported."))
+    end
+    d === 0 ? _round_digits0(x, r) : _round_digits(x, r, d)
+end
+function _round_digits(x::X, r::RoundingMode, d::Int) where {T, f, X <: FixedPoint{T,f}}
+    log10_2 = 0.3010299956639812
+    d > floor(Int, ((f + 1) * log10_2)) && return x
+    r = round(float(x), r, digits=d)
+    typemin(X) - eps(X)/2 <= r < typemax(X) + eps(X)/2 || throw_converterror(X, r)
+    clamp(r, X)
+end
+
+trunc(x::X) where {X <: FixedPoint{<:Unsigned}} = floor(x)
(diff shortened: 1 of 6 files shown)


kimikage:
## Benchmark
```julia
using ColorTypes, BenchmarkTools
image = rand(RGB{N0f8}, 1000, 1000);

function quantize(x::N0f8)
    floor(x, digits=1)
end
function quantize(color::AbstractRGB)
    mapc(quantize, color)
end


function quantize_f(x::N0f8)
    floor(float(x), digits=1) % N0f8
end
function quantize_f(color::AbstractRGB)
    mapc(quantize_f, color)
end
```

```julia
julia> @benchmark quantize.($image)
BenchmarkTools.Trial:
  memory estimate:  2.86 MiB
  allocs estimate:  2
  --------------
  minimum time:     1.504 ms (0.00% GC)
  median time:      1.532 ms (0.00% GC)
  mean time:        1.731 ms (10.89% GC)
  maximum time:     5.461 ms (71.47% GC)
  --------------
  samples:          2888
  evals/sample:     1

julia> @benchmark quantize_f.($image)
BenchmarkTools.Trial:
  memory estimate:  2.86 MiB
  allocs estimate:  2
  --------------
  minimum time:     18.977 ms (0.00% GC)
  median time:      19.034 ms (0.00% GC)
  mean time:        19.298 ms (1.17% GC)
  maximum time:     22.876 ms (16.72% GC)
  --------------
  samples:          260
  evals/sample:     1
```

```julia
n4f4  = rand(N4f4,  1000);
n0f8  = rand(N0f8,  1000);
n0f16 = rand(N0f16, 1000);

q3f4  = rand(Q3f4,  1000);
q0f7  = rand(Q0f7,  1000);
q0f15 = rand(Q0f15, 1000);
```

```julia
julia> @btime trunc.($n4f4,  digits=1); # not optimized
  13.199 μs (1 allocation: 1.06 KiB)

julia> @btime trunc.($n0f8,  digits=1);
  161.658 ns (1 allocation: 1.06 KiB)

julia> @btime trunc.($n0f8,  digits=2);
  158.163 ns (1 allocation: 1.06 KiB)

julia> @btime trunc.($n0f16, digits=2); # not optimized
  15.100 μs (1 allocation: 2.13 KiB)

julia> @btime trunc.($q3f4,  digits=1);
  221.577 ns (1 allocation: 1.06 KiB)

julia> @btime trunc.($q0f7,  digits=1);
  126.801 ns (1 allocation: 1.06 KiB)

julia> @btime trunc.($q0f7,  digits=2);
  222.530 ns (1 allocation: 1.06 KiB)

julia> @btime trunc.($q0f15, digits=2); # not optimized
  5.767 μs (1 allocation: 2.13 KiB)
```

kimikage:
This changes the behavior as a result, but I think this is rather a bug fix. So, I will merge this.

PR #238, opened 2020-12-16
Use GitHub Actions for unit testing on Arm arch.

### LINKED ISSUE #237: Migration of CI for ARM
We've aggregated the major CI jobs into GitHub Actions, in terms of speed and UI integrity (PR #204).
However, the GitHub-hosted runners (i.e. non-self-hosted runners) do not support ARM architecture. For this reason, we currently use Travis CI for arm64 (aarch64).

We currently have two problems with Travis CI. (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/pull/204#issuecomment-722294424)
One is that FixedPointNumbers.jl uses the old "travis-ci.org" and needs to be migrated to "travis-ci.com" by the end of this year.
The other is that Travis CI's pricing plan has been changed. (https://discourse.julialang.org/t/travis-ci-changes-pricing-plans/49528)

I tried out the [Drone Cloud](https://cloud.drone.io/) as a free CI service that supports the ARM architecture. 
(cf. https://cloud.drone.io/kimikage/ColorBlendModes.jl/5)

Are there any other good tools out there?

kimikage:
The test runs on a docker container with QEMU.
This also updates TagBot.yml.

Closes #237


.github/workflows/UnitTestArm.yml
```diff
@@ -0,0 +1,38 @@
...
+        distro: [ubuntu20.04]
+        arch: [aarch64]
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: uraimo/run-on-arch-action@v2.0.7
+        with:
+          arch: ${{ matrix.arch }}
+          distro: ${{ matrix.distro }}
+          githubToken: ${{ github.token }}
```
johnnychen94, line 26: is this for code coverage upload?

kimikage: This is used for pushing the cache images. However, I am not sure about its necessity.

.github/workflows/UnitTestArm.yml
```diff
@@ -0,0 +1,36 @@
...
+      - master
+  pull_request:
+  workflow_dispatch:
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: ['1.0', '1', 'nightly']
```
johnnychen94, line 15: maybe just `1` to save CI build time? I'm not sure.

johnnychen94 (approved):
The config file looks good to me. We might hit with some unrelated GA/docker issue. I'd suggest trying it again a few hours later or tomorrow.

.github/workflows/UnitTestArm.yml
kimikage: I was expecting python3 to be included in the docker images, but it appears not to be. Jill.py is a great tool if we already have python installed, but it has too large a footprint to be used just for installing Julia. :sweat_smile:
However, the original jill.sh does not support arm at the moment.

johnnychen94: > Jill.py is a great tool if we already have python installed, but it has too large a footprint to be used just for installing Julia.

Indeed. Unfortunately, I'm not a bash wizard to add the functionality I wanted (local release registry), so I have to use Python to do it.

I use this to set up CI in the self-hosted gitlab in my school servers, and with properly set up apt/jill/pip mirrors, I can make sure most network traffics are limited in the LAN network.

johnnychen94:
😱

johnnychen94:
Perhaps temporarily remove `.github/workflows/UnitTest.yml`?

kimikage:
I spent half the day on this silly typo. :joy:
`shell: bin/bash` ->`shell: /bin/bash`

It should not be necessary to download julia in the container. However, `setup-julia` also does not support arm.

kimikage:
> test (1.0, ...     27m 29s
> test (1, ...       43m 18s
> test (nightly, ... 26m 2s

Although I had expected it, it seems too slow. :sob:
However, compared to [Travis CI](https://travis-ci.org/github/JuliaMath/FixedPointNumbers.jl/builds), GitHub Actions has a higher parallelism, so the total time does not seem to be drastically different.
Umm....

.github/workflows/UnitTestArm.yml
kimikage: Since those tests are run in parallel, I do not think reducing the targets will have much of an effect on speeding up the total process.

kimikage:
There is some room for improvement, but I'd like to try this out first.

Thank you, Travis CI, for all your support!

johnnychen94:
Is the slow CI time a container issue, or just FixedPointNumbers has bad performance in Arm?

kimikage:
I think the speed depends on the CI environment. At least the Travis CI tests which ran on the Arm machine were 2-3 times faster per configuration (i.e. julia-version).

I don't have a fast Arm machine, though.

PR #255, opened 2022-07-07
using Requires for Statistics

johnnychen94:
Because Statistics is no longer part of the Julia sysimage, the package loading
time of FixedPointNumbers is then donimated by Statistics. This commit uses
Requires to conditionally add the needed method without making more one-line
package. This, of course, adds some additional overhead but I do believe it's
currently the best tradeoff until we have first-class support on conditional
package loading.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,9 +4,11 @@ version = "0.8.4"
 
 [deps]
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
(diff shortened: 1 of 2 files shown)


timholy (approved):

johnnychen94:
I'm merging this, but we won't be able to make a new release because many breaking changes have been introduced since the last release. This is why I've tagged a dev version #258 to avoid a breaking patch release by accident.

PR #256, opened 2022-07-07
fix test for Julia > 1.7

johnnychen94:


johnnychen94:
Looks like I'm trying to fix something already fixed :)

PR #277, opened 2024-02-12
Enable package extension for Statistics

hyrodium:
This PR reverts #255.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,8 +13,10 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 FixedPointNumbersStatisticsExt = "Statistics"
 
 [compat]
+# Update this version specifier when Statistics.jl v1.11.2 is released.
+# https://github.com/JuliaStats/Statistics.jl/issues/165
+Statistics = "< 1.11.2"
 julia = "1"
-Statistics = "1 - 1.11.1"
 
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,12 @@ version = "0.9.0-dev"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
+[weakdeps]
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
+[extensions]
+IntervalSetsStatisticsExt = "Statistics"
+
 [compat]
 julia = "1"
 
--- a/Project.toml
+++ b/Project.toml
@@ -14,6 +14,7 @@ FixedPointNumbersStatisticsExt = "Statistics"
 
 [compat]
 julia = "1"
+Statistics = "1 - 1.11.1"
 
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
--- a/Project.toml
+++ b/Project.toml
@@ -10,14 +10,15 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [extensions]
-IntervalSetsStatisticsExt = "Statistics"
+FixedPointNumbersStatisticsExt = "Statistics"
 
 [compat]
 julia = "1"
 
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Documenter", "Test"]
+test = ["Documenter", "Statistics", "Test"]
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -629,4 +629,8 @@ if VERSION >= v"1.1" # work around https://github.com/JuliaLang/julia/issues/341
     _precompile_()
 end
 
+if !isdefined(Base, :get_extension)
+    include("../ext/FixedPointNumbersStatisticsExt.jl")
+end
+
 end # module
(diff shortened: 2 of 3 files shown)


kimikage:
This seems more elegant than the use of Requires.jl with respect to Stdlib.

On the other hand, in the first place, this workaround uses a non-public function `Statistics._mean_promote`.
Until now, this has been acceptable because the version of Statistics.jl was synchronized with Julia's version upgrades.

Seems to me we need to revisit the workaround itself.
Of course, that should be separate from this PR.

kimikage:
This change is not so complicated, and the intent is clear.
If we decide to merge this PR, however, it might be easier to track if the simple `revert` commit was made in master.
This is just a little reminder for the hygiene of commit-graph, because PR #278 also makes changes to Project.toml.

johnnychen94:
Sorry, I missed this PR.

For the very fundamental packages such as FixedPointNumbers, keeping a broad range of compatibility is necessary. FixedPointNumbers is still compatible with Julia 1.0, which is great.

Thus, we still need to preserve the old codes, still need to install Requires (but conditionally loading it depending on the actual Julia versions).

Ref: https://pkgdocs.julialang.org/v1/creating-packages/#Backwards-compatibility

kimikage:
@johnnychen94 You are absolutely right.
We might need Requrires.jl (perhaps for CheckedArithmetic?).

But for the case of Statistics, it has been a stdlib and should work as before. (Conversely, the "new” Statistics.jl basically won't be introduced into the old julia.)

johnnychen94:
> Adding extension support is a good change, but it requires Julia 1.9 to work.

Sorry, I misunderstand the code. The current version **_is_** compatible with old Julia versions by falling back to a direct `include` rather than the lazy `@require`.

My two cents is to still use the lazy `@require` because it's what we expected in #255.

src/FixedPointNumbers.jl
```diff
@@ -625,20 +624,8 @@ if VERSION >= v"1.1" # work around https://github.com/JuliaLang/julia/issues/341
...
-        end
-    end
-else
-    import Statistics
-    if isdefined(Statistics, :_mean_promote)
-        # https://github.com/JuliaMath/FixedPointNumbers.jl/pull/183
-        Statistics._mean_promote(x::Real, y::FixedPoint) = Treduce(y)
-    end
+if !isdefined(Base, :get_extension)
+    include("../ext/FixedPointNumbersStatisticsExt.jl")
```
kimikage, line 628: Although this is in keeping with PR #255 manner, I would prefer that the `precompile` be at the end of the module (except for the `__init__()`).

kimikage:
@hyrodium Could you `rebase` (and tweak) this?

Project.toml
```diff
@@ -4,16 +4,21 @@ version = "0.9.0-dev"
...
 julia = "1"
 
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Documenter", "Test"]
+test = ["Documenter", "Statistics", "Test"]
```
kimikage, line 24: I don't fully understand how Package Extension works, but this package still hardly `[dep]`ends on `Statistics` for the time being, so I guess it is not needed for `[extras]` and `test`.

johnnychen94 (approved):

Project.toml
```diff
@@ -4,16 +4,21 @@ version = "0.9.0-dev"
...
 
+[weakdeps]
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
+[extensions]
+FixedPointNumbersStatisticsExt = "Statistics"
+
 [compat]
-Requires = "1"
 julia = "1"
```
kimikage, line 16: If we follow the principle, `Statistics` should be in `[compat]`.
This is especially the case where even patch versions need to be included (e.g. `Statistics = "1 - 1.11.1"`).

kimikage (commented):
It is hard to determine where in the trade-off we should compromise.

kimikage:
cf. https://github.com/JuliaStats/Statistics.jl/issues/165

kimikage:
@hyrodium Thank you for rebasing.
Is this ready to merge? Or do you plan to revise it (regarding my comments)?
I will merge this PR once I get the go-ahead.

Project.toml
hyrodium: Fixed in https://github.com/JuliaMath/FixedPointNumbers.jl/pull/277/commits/6c2a68cc9513d092be75289e840daac58fc3c2e0.

src/FixedPointNumbers.jl
hyrodium: Fixed in https://github.com/JuliaMath/FixedPointNumbers.jl/pull/277/commits/c4eee40f2aa33258e150259de49468bf73c20838.

hyrodium: Actually, I'm not familiar with `_precompile_()` function. Perhaps the current code should be replaced with [PrecompileTools.jl](https://github.com/JuliaLang/PrecompileTools.jl).

Project.toml
kimikage: To my surprise, it seems that the hyphen notation is not supported for julia v1.0.
Sorry for misleading you.

src/FixedPointNumbers.jl
kimikage: Yes, it is from the SnoopCompile.jl days.
I think replacing before releasing the new version is a good idea.

Project.toml
hyrodium: Yeah, `< 1.11.2` could be a solution (cf. [Pkg.jl v1.0 documentation](https://pkgdocs.julialang.org/v1.0/#Inequality-specifiers-1)). However, `"1 - 1.11.1"` looks better to me because it has a lower bound. Is there any chance to drop support for Julia 1.0?

kimikage: Since Statistics.jl is stdlib, the lower bound is practically meaningless.
I don't intend to continue actively supporting v1.0, but I don't think it has caused any critical problems yet.

kimikage: Also, this is very trivial, the uppercase "S" of Statistics precedes the lowercase "j" of julia.

hyrodium: Thanks for the comments! I have updated the compat bound for Statistics.jl

kimikage:
@hyrodium Thank you for addressing this issue!

PR #278, opened 2024-04-05
Use `StableRNG` in tests

kimikage:
Although reproducibility was not guaranteed, the output of `MersenneTwister` seems to have changed.

Note that `rand(StableRNG(1234), T)` is not a test for the generated random numbers, but for passing an RNG to the argument of `rand`.
On the other hand, since `StableRNGs` are not part of stdlib, I added the tests for the default RNG.


kimikage:
xref: https://github.com/JuliaMath/FixedPointNumbers.jl/pull/277#issuecomment-2040851472

johnnychen94 (approved):
A harmless change to me.

It's so glad to see you well @kimikage !

PR #287, opened 2024-04-14
Add workaround for julia-invalidations

kimikage:
This is very ugly, but it is better than not being able to use the check.

xref: https://github.com/julia-actions/julia-invalidations/issues/17

Another possible method is to use git, but directly overwriting ”Project.toml” is simpler and somewhat more reliable.


kimikage:
BTW, the output does not seem very practical.

Issue #288, opened 2024-04-26
Problem with  `unsafe_trunc`  returning arbitrary values


kimikage:
I have dealt with the `rem` (more specifically `unsafe_trunc`) problem on ARM, mainly with PR #136, #223.
However, this code seems to cause bad problems under certain conditions. (cf. https://github.com/JuliaLang/julia/issues/54264)

That is only likely to happen on ARM, but it can also happen on x86_64 (e.g., -1 % N0f8).

kimikage:
It is better to switch to a safer implementation, even if it means sacrificing some speed. Also, it would be better to backport the fix to v0.8.

kimikage:
Fortuitously, this issue number is the same as the issue number in ColorTypes.jl.😇
https://github.com/JuliaGraphics/ColorTypes.jl/issues/288

kimikage:
This was fixed by #291, but we need to backport it to v0.8 and release it.

PR #289, opened 2024-04-27
Add test for undefined behavior of `unsafe_trunc`

kimikage:
This is in preparation for fixing issue #288.
The actual modification will be done in a separate PR.


kimikage:
Fine, everything is broken except the nightly build!

In my defense, this was (potentially) broken before I changed it and would have been even more broken if I hadn't.

PR #293, opened 2024-04-30
[RFC] Backports for v0.8.5

kimikage:
### To be backported
- [x] `NotBiggerThanInt64`, `nbitsint`, `showtype` (`#189`)
- [x] #205
- [x] #208
- [x] `widen1(T::Type)`, tests (`#207`)
- [x] #206
  - Needs #278
- [x] (partial) #214
- [x] #212 
- [x] #218
- [x] (partial) #213
- [x] (partial) #220
- [x] #223
  - Needs #291 
- [x] (partial) #224
- [x] tests (`#222`)
- [x] (partial) #226
- [x] #229
- [x] #231
- [x] move `trunc` (`#235`)
- [x] #238
- [x] #233
- [x] `QuoteNode`,`return nothing`, tests(`#243`)
- [x] #244
- [x] #248
- [x] #253
- [x] #254
- [x] #257
- [x] #263
- [x] #270
- [x] #276
- [x] #279
- [x] #280
- [x] #277
- [x] #278
- [x] #283
- [x] #286
- [x] #290
- [x] #289
- [x] #291
- [x] #295

### Not to be backported

| PR |Reason / Note |
|:--:|:-------------|
|#189|💥Breaking (partially backported✅)|
|#207|💥Breaking (partially backported✅)|
|#190|Too many changes|
|#216|💥Breaking|
|#217|Dependent on #190|
|#222|💥Breaking (partially backported✅)|
|#225|Replaced by #283|
|#230|Dependent on #190|
|#235|💥Breaking (partially backported✅)|
|#236|💥Breaking|
|#243|💥Breaking (partially backported✅)|
|#249|Reverted|
|#258|💥for v0.9|
|#255|Reverted|
|#262|Not needed|
|#267|Not needed|
|#266|Not needed|
|#265|Not needed|
|#269|Not needed|
|#273|Not needed|
|#275|Not needed|
|#281|Not needed|
|#259|Not needed|
|#284|Not needed|
|#282|Not needed|
|#287|Not needed|

PR #222, opened 2020-08-25
Add checked arithmetic for `/`

kimikage:
This also changes the implementation of `/` for `Fixed`. The new `/` also checks for overflow.

`/` (`checked_fdiv`) throws `DivideError` for division by zero and `OverflowError` for overflow.
```julia
julia> 0.5Q0f7 / 0Q0f7 # v0.8.4 and this PR
ERROR: DivideError: integer division error

julia> 0.5N0f8 / 0N0f8 # v0.8.4
ERROR: ArgumentError: Normed{UInt8,8} is an 8-bit type representing 256 values from 0.0 to 1.0; cannot represent Inf

julia> 0.5N0f8 / 0N0f8 # this PR
ERROR: DivideError: integer division error
```
```julia
julia> -1Q0f7 / -1Q0f7 # v0.8.4
-1.0Q0f7

julia> -1Q0f7 / -1Q0f7 # this PR
ERROR: OverflowError: -1.0Q0f7 / -1.0Q0f7 overflowed for type Q0f7

julia> 1N0f8 / 0.5N0f8 # v0.8.4
ERROR: ArgumentError: Normed{UInt8,8} is an 8-bit type representing 256 values from 0.0 to 1.0; cannot represent 1.9921874

julia> 1N0f8 / 0.5N0f8 # this PR
ERROR: OverflowError: 1.0N0f8 / 0.502N0f8 overflowed for type N0f8
```

PR #296, opened 2024-05-01
Version 0.8.5

kimikage:
Backporting (PR #293) is largely complete.
Division `/` is inconsistent, but this is due to the existing implementation. （It should be improved in v0.9. #222）

I will now check on the impact on the downstream packages.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,15 +1,34 @@
 name = "FixedPointNumbers"
 uuid = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
-version = "0.8.4"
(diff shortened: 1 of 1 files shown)


kimikage:
It might be a good idea to review the precompilation settings, but I am very tired.

kimikage:
I ran out of steam trying to maintain the tests of downstream packages such as ColorTypes.jl and Colors.jl.
It will take a little more time. 🤕

kimikage:
The `[compat]` specification for Statistics seems to cause compatibility problems in downstream packages, ironically.

PR #320, opened 2026-03-11
Backport "Remove assertions in "src/precompile.jl" (#254)"

Kyjor:
- Cherry-picked commit #59ee94b
- Bumped version number
- Removed macos-13 os from the matrix. No longer supported by actions.
- Updated arm action version, as the current on no longer works.

### DIFF
--- a/.github/workflows/UnitTest.yml
+++ b/.github/workflows/UnitTest.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         julia-version: ['1.0', '1.6', '1', 'nightly']
-        os: [ubuntu-latest, windows-latest, macos-13]
+        os: [ubuntu-latest, windows-latest]
         julia-arch: [x64]
         include:
           - os: ubuntu-latest # only test one 32-bit job
--- a/.github/workflows/UnitTestArm.yml
+++ b/.github/workflows/UnitTestArm.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         julia-version: ['1.0', '1.6', '1', 'nightly']
         os: [ubuntu-latest]
-        distro: [ubuntu_latest]
+        distro: [ubuntu24.04]
         arch: [aarch64]
 
     steps:
@@ -51,8 +51,7 @@ jobs:
           mkdir -p /home/runner/work/julia/
           tar -xf /tmp/julia-aarch64.tar.gz --strip-components=1 -C /home/runner/work/julia/
           rm /tmp/julia-aarch64.tar.gz
-
-      - uses: uraimo/run-on-arch-action@v2.7.2
+      - uses: uraimo/run-on-arch-action@v3
         name: Unit Test
         with:
           arch: ${{ matrix.arch }}
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FixedPointNumbers"
 uuid = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
-version = "0.8.5"
+version = "0.8.6"
 
 [deps]
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -6,27 +6,27 @@ function _precompile_()
     realtypes = (Float16, Float32, Float64, Int)     # types for mixed Normed/Real operations
     for T in normedtypes
         for f in (+, -, abs, eps, rand)       # unary operations
-            @assert precompile(Tuple{typeof(f),T})
+            precompile(Tuple{typeof(f),T})
         end
-        @assert precompile(Tuple{typeof(rand),T,Tuple{Int}})
-        @assert precompile(Tuple{typeof(rand),T,Tuple{Int,Int}})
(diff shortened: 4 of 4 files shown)


heyx3:
Can I ask, Why do precompile statements fail when using Package Compiler in the first place? I can't find much info on this online.

Also, do more changes need to be made to this PR to satisfy the CodeCov bot or are we just waiting for somebody to override it?

Kyjor:
@heyx3  honestly, I'm not sure about _why_ they fail, but it seems like there was a bigger discussion about removing these previously: https://github.com/JuliaGraphics/ColorTypes.jl/pull/232#issuecomment-790993554. Seems like it was more about performance? 

Also I didn't add any coverage as it doesn't seem to make any sense (at least as far as I understand) to test precompilation. If needed, I can do it though. At this point I'm just hoping someone who manages this repo sees this relatively soon. 

I've had to [remove a dependency](https://github.com/Kyjor/JulGame.jl/pull/96/commits/c4d097c13fd72a0ccad9c97112ce49bd209ddf4e) from my game engine temporarily so I can continue to build my game. Not a big deal for me for now, but I'm sure a few more people are depending on this downstream than just us

heyx3:
Bump. This is a small change, and in the meantime any projects which happen to bring this package in as a sub-dependency are totally unable to use Package Compiler.

bjarthur:
this repo has an unusual (for julia packages) branching structure.  normally a release is tagged on the master branch and subsequent development then continues on the master.  so no release branch is created.  but @kimikage for some reason made a v0.8.5 branch from commit cab78d7, made three commits and then tagged a release.  subsequent then work continued on master, omitting those three commits.  see

https://github.com/JuliaMath/FixedPointNumbers.jl/commits/v0.8.5/

https://github.com/JuliaMath/FixedPointNumbers.jl/commits/master/?after=d70026a230c318cb0ed83c5f6bb187a6bdd3cd8a+69

there has been a lot of work on master since 0.8.5.  to include removing the asserts on those precompiles:

https://github.com/JuliaMath/FixedPointNumbers.jl/blob/master/src/precompile.jl

so i think all that needs to be done here, is decide whether the master branch is stable enough to tag a new release.  and if so, decide on a version number.


@johnnychen94 @ViralBShah @vchuravy

ViralBShah:
How about we tag the latest master as 0.8.6?

Should this PR be merged (even though some of the things have been implemented)?

bjarthur:
the master branch of Project.toml has 0.9.0-dev as the version.

i worry that @kimikage made a version branch because he didn't think the master branch was stable.  tests do pass for me on master though.  do we have any way to contact him?

ViralBShah:
I don't have any other way of contacting except for pinging here. Maybe we just go to 0.9.0 then? In that case, it may also be good to push the julia compat up to 1.10.

I do not know enough about the internals here, beyond helping with releases and maintenance.

heyx3:
> made a v0.8.5 branch from commit cab78d7, made three commits and then tagged a release.

Given that, and given that it's fixed in main, why not just merge this into the 0.8.5 branch to patch that specific version?

ViralBShah:
It seems the goal was to have a separate branch for 0.8 fixes, and master live on 0.9. Fair point - that we could just continue the fixes for 0.8 on that branch and tag 0.8.6.

Is someone chiming in here happy to continue the development to release 0.9? It should be easy to see what the differences are between 0.8.x and 0.9-dev.

bjarthur:
from the commit at which master branched from v0.8.5, this is what was changed for v0.8.5:

git diff cab78d7c4aa31673cc6f9f7f9de8aaddef2e8854 c372941791804d35e45b03615edd1dcb86a09da8

and this is what changed on master:

git diff cab78d7c4aa31673cc6f9f7f9de8aaddef2e8854 cadbbc1b1f0ca339bd989e4b4fc4a4d800c78b3d

almost everything in the former is in the latter, but there is quite a bit in the latter than is not in the former.

i've tried to contact @kimikage on discourse.

ViralBShah:
They haven't been active for 2 years, and there was a personal update in https://discourse.julialang.org/t/pre-rfc-ann-color-related-packages-yuvcolortypes-cmykcolortypes-colorprofiles-hdrcolortypes-hdrimages/113032.

Perhaps it is safe to assume that we won't hear. It would be great if someone here would want to take ownership, and I'm happy to help facilitate new maintainers.

bjarthur:
@inkydragon @hyrodium @ranocha @timholy @johnnychen94

you all have contributed commits to FixedPointNumbers.jl since the last release was tagged.  having you been using the master branch in production?  i ask because we are curious whether it is stable enough to tag a new release.

for context, see the [thread above](https://github.com/JuliaMath/FixedPointNumbers.jl/pull/320#issuecomment-4335370067) about the current release being a side branch instead of a commit on master.  that @kimikage chose to make this side branch, instead of tagging master, makes me worry that master is not stable.  hence my question to you.

ranocha:
> you all have contributed commits to FixedPointNumbers.jl since the last release was tagged. having you been using the master branch in production? i ask because we are curious whether it is stable enough to tag a new release.

No, I have not.

ViralBShah:
I suggest that since we do not have a view on the master, how about get things merged in the current way things are done (on the branch rather than main) and at least release a new version that fixes the current issue?

PR #324, opened 2026-05-21
Add JET testing to the 0.8 branch

PatrickHaecker:


kimikage:
What specific problem is this intended to solve?
To what extent does this offer advantages over running the test locally?

I'm not opposed to using JET, but providing an explanation might make it easier to gain the support of other maintainers.

PR #325, opened 2026-06-02
bump minor version

bjarthur:
@kimikage had described the master branch as "dev".  tests pass though.  so we'll release it into the wild.  let us know if any downstream dependencies break, and use v0.8.6 in the meantime.


bjarthur:
from google antigravity:

  Here are the key breaking or behavioral changes you should be aware of when upgrading from 
  v0.8.5  to  v0.9 :
  ──────
  ### 1. Silent Overflow in  Normed  Multiplication (Breaking Change)

  • In v0.8.5: The  *  operator on  Normed  types was explicitly overridden to use checked   
  arithmetic ( checked_mul ). If a multiplication overflowed, it immediately threw an        
  OverflowError .
  • In v0.9: This override was removed. Multiplication on all fixed-point types (including   
  Normed  types) now defaults to wrapping arithmetic ( wrapping_mul ).
  • Impact: Overflows during multiplication will now silently wrap around instead of throwing
  errors. Any code relying on  OverflowError  safety checks during multiplication will fail
  silently.
  ──────
  ### 2. Printing & String Representations (Breaking Change)

  • In v0.8.5:  print(x)  fell back to  show(x)  and output the number with its type suffix
  (e.g.,  print(q0f7)  output  "-0.672Q0f7" ).
  • In v0.9:  print  and  show  were decoupled.  print(x)  now prints only the clean numeric 
  value (e.g.,  "-0.672" ), while  show  handles the type suffix and type representation.
  • Impact: Any user code that performs string parsing on printed representations, relies on
  string(x) , or uses string interpolation ( "$x" ) and expects the type suffix (like  Q0f7 
  or  N0f8 ) to be present will break.
  ──────
  ### 3. Range Length Behavior (Behavioral Change)

  • In v0.8.5: Calling  length  on step ranges of fixed-point numbers that overflowed would
  return incorrect dummy values (e.g.  Base.unsafe_length(...) == -1  or  0 ).
  • In v0.9: These overflowing step ranges now explicitly throw an  OverflowError .
  • Impact: While this is a safety correction, it changes the execution flow for any code
  that previously processed or absorbed those invalid range length values.
  ──────
  ### 4. Deprecation Warnings (Potential Future Breaks)

  If users use internal or fallback APIs, they will now receive deprecation warnings in
  stderr:

  •  typechar : Deprecated in favor of  type_prefix  (which returns symbols like  :Q  instead
  of character  'Q' ). Calling  typechar  now warns the user.
  •  floattype  Fallback: Fallbacks for  floattype  on custom real types now emit a warning
  alerting the user that a future release will throw a  MethodError  if a type               
  <:AbstractFloat  cannot be returned.

PR #327, opened 2026-06-17
Support parsing

PatrickHaecker:
Support `Base.parse` and `Base.tryparse` for `Normed` and `Fixed`.

The PR builds on #324 and targets FixedPointNumbers.jl in v0.8.


kimikage:
The development branch has been moved to v0.9.

I don’t think adding a parser is a bad idea, but we need to discuss which formats to accept.

If there are ambiguities or limitations in the specification, or if it is essentially a reimplementation of `parse(Float64, x)`, I think it’s better to leave that matter to downstream packages or end users.

